PR #13, opened 2013-02-10
Add dependency on generated messages

arhachey:
Catkin requires explicit enumeration of dependencies on generated messages.
Add this declaration to properly flatten the dependency graph and force Catkin
to generate geometry_msgs before compiling image_geometry.


vrabaud:
thx for the patch ! Will make a new release with it.

Issue #27, opened 2013-11-22
Image Message step handling broken


jensenb:
The cv_bridge handling of the image step size field of an image message is broken. Introduced in ee10a9dc9fd20145d04c47e8c2fe50a8a6f4c3a1 there is now an invariant when converting ROS Image Messages to `cv::Mat` instances where the step size has to be equal to the image element size \* the number  of channels \* the image width. If this invariant is not fulfilled an exception is thrown, crashing rqt_image_view for example. This is requirement is not listed on the wiki page, and is not listed in the Image Message documentation: http://docs.ros.org/api/sensor_msgs/html/msg/Image.html. Aligned Image Messages are thus no longer supported.

I see no reason why the cv_bridge step handling has to forbid aligned images, when the underlying OpenCV API has no problem with this.

jensenb:
I can submit a pull request if that would help

vrabaud:
I will have time in a few hours (3/4). The pull request is more than welcome but please, also add a test for it.

vrabaud:
that got fixed in #28

PR #28, opened 2013-11-22
Fixed issue with image message step size

jensenb:
This should fix issue #27, together with a fitting unit tests.


vrabaud:
ok, thx. Looks good. Just to be picky (it' fun ! :) ):
- can you please merge the two tests. I know, it's better to separate tests but most of the code is the same and I'd like to see another similar test in there (see below)
- can you not do a push_back and simply do a resize of .data (just to have less code)
- can you also add a case where the step is equal to the width

Thx ! I'll release it tonight (I also have some numpy code in there that works and needs to be released)

jensenb:
Here you go.

vrabaud:
That looks perfect ! Thank you very much for your contribution and responsiveness. Releasing now !

PR #29, opened 2013-11-23
Find NumPy include directory

jensenb:
The NumPy include directories are not set, causing build errors on OS X. I noticed this already while doing my previous pull request, but thought it was related to a misconfiguration on my system. Really I don't know how this builds on Ubuntu, the devs must place the NumPy headers inside of the Python include directory.

### DIFF
--- a/cv_bridge/src/CMakeLists.txt
+++ b/cv_bridge/src/CMakeLists.txt
@@ -7,10 +7,27 @@ install(TARGETS ${PROJECT_NAME} DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION})
 
 # add a Boost Python library
 set(Python_ADDITIONAL_VERSIONS 2.7)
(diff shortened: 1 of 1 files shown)


vrabaud:
thx ! That could work on Windows too. I found a good FindNumPy.cmake here too:
http://svn.lam.fr/repos/unsio/trunk/py/cmake/FindNumPy.cmake
Maybe that could/should go inside catkin like the gtest and Eigen ones.

jensenb:
Yeah it would probably be cleaner to have separate FindNumPy.cmake module. The version version I proposed here is very similar to your linked version, with the difference that the other version also checks for the NumPy version, but I don't think that is currently necessary.

vrabaud:
ok, I made a pull request:
https://github.com/ros/cmake_modules/pull/8

PR #30, opened 2014-01-17
Added operator= for the PinholeCameraModel.

TobiasBaer:
Added the operator= for the PinholeCameraModel. I am not sure if the
PinholeCameraModel needs to have a destructor, too. To follow the
'rule of three' it should actually have one.

[1] http://answers.ros.org/question/118911/image_geometrypinholecameramodel-operator-missing/


vrabaud:
ok, that breaks the ABI but it's worth having. Could you also add the assignment operator to StereoCameraModel, thx.

PR #31, opened 2014-01-22
fix python 3 error at configure time

dirk-thomas:
This only addresses a configure time problem with Python 3. Compilation still fails when using Python 3. This should also be addressed in the very near future.


vrabaud:
Thx, I'll accept it for now but the long term solution is to merge https://github.com/ros/cmake_modules/pull/8

PR #32, opened 2014-02-05
fixed Python rectifyImage implementation in PinholeCameraModel

mdouskos:
The implementation of rectifyImage(...) method of PinholeCameraModel in Python was not equivalent to the C++ one. As it was based on InitUndistortMap(...) function of OpenCV, it only undistorted the image without using the rotation and the new camera matrix specified in the camera model. I changed it with the InitUndistortRectifyMap(...) function of OpenCV which considers also these transformations.


vrabaud:
nice one, thx !

PR #37, opened 2014-05-19
pinhole_camera_model: fix implicit shared_ptr cast to bool for C++11

xqms:
In C++11 boost::shared_ptr does not provide the implicit bool conversion
operator anymore, so make the cast in pinhole_camera_model.h explicit.
That does not hurt in older C++ standards and makes compilation with C++11
possible.

### DIFF
--- a/image_geometry/include/image_geometry/pinhole_camera_model.h
+++ b/image_geometry/include/image_geometry/pinhole_camera_model.h
@@ -257,7 +257,7 @@ public:
   /**
    * \brief Returns true if the camera has been initialized
    */
-  bool initialized() const { return cache_; }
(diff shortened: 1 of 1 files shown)


raphaelmemmesheimer:
+1 

this also affects C++0x, at least with the g++ 4.8.2 compiler shipped with ubuntu 14.04. The same code was working with C++0x and g++ 4.7.3 and ubuntu 13.04.
An other solution would be `return cache_.get();`

vrabaud:
sounds good, thx !

vrabaud:
I'll release that as soon as #38 is closed (except if you consider it more urgent).

PR #38, opened 2014-05-23
Correct dependency from non-existent package to cv_bridge

130s:
I'm in the middle of getting `opencv_tests` run to see how this package is usable (or not...what for is this package included in `vision_opencv`?).

### DIFF
--- a/cv_bridge/python/cv_bridge/core.py
+++ b/cv_bridge/python/cv_bridge/core.py
@@ -40,7 +40,7 @@ import warnings
 
 class CvBridgeError(TypeError):
     """
-    This is the error raised by :class:`opencv_latest.cv_bridge.CvBridge` methods when they fail.
+    This is the error raised by :class:`cv_bridge.CvBridge` methods when they fail.
     """
     pass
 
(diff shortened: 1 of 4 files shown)


vrabaud:
not sure. broadcast.py and source.py should be rostests don't you think ? The face one is just an example.

130s:
Do you mean it would make more sense to move broadcast.py and source.py to
cv_bridge? If so I'll edit my PReq.

vrabaud:
they would yes, as rostest. I never looked at that test package thinking it was just standalone apps, but they actually are valid rostest (for those two files at least)

130s:
Ok, then I'll move those python files to `cv_bridge` package, test, commit and notify here again.

vrabaud:
Thx ! It's totally something I should have done but thx for your help !

130s:
Not a problem, I need unit tests for this package anyway.

Concerns are:
- 1) I can't work on this until later this week (so that this ticket would block ([this](https://github.com/ros-perception/vision_opencv/pull/37#issuecomment-44719663)). I think testing is not urgent at all so I suggest you go ahead release new version.
- 2) I realized that dependency has changed in `Indigo` branch while my local working patch is against Hydro (ie. `groovy-devel` branch). Are you ok with me re-opening a pull request against `groovy-devel` first then later synchronize appropriate commits into `Indigo`? Or any better suggestion?

vrabaud:
we can wait a bit, and you can do your PR on groovy-devel, no worries: just try to have ROS specific version commits apart.

vrabaud:
I had totally overlooked those files: they are not tests, just utilities to help you debug, I added a README there.

130s:
Yeah, I realized that too. So what I ended up doing is to write almost a new `unittest` python file that tests methods in `CVBridge` class. As said I'll open a new PR against `groovy-devel` first.

PR #43, opened 2014-07-18
Fix #42

### LINKED ISSUE #42: cv_bridge: crashes for single channel images
At [core.py:157](https://github.com/ros-perception/vision_opencv/blob/indigo/cv_bridge/python/cv_bridge/core.py#L157) number of channels is retrieved from the third place of the `shape` variable, however for the single channel images the `shape` has just two fields.

liborw:
This should fix the problem with single channel images.


vrabaud:
thx !

PR #45, opened 2014-09-06
Using Opencv3.0.0

edgarriba:


cv_bridge/src/module.cpp
```diff
@@ -227,16 +229,16 @@ int convert_to_CvMat2(const PyObject* o, cv::Mat& m)
 
     if( m.data )
     {
-        m.refcount = refcountFromPyObject(o);
+        //m.refcount = refcountFromPyObject(o);
```
vrabaud, line 232: why is that commented out ?

cv_bridge/src/module.cpp
```diff
@@ -145,7 +147,7 @@ void NumpyAllocator::deallocate( int* refcount, uchar* datastart, uchar* data )
 }
 
 // Declare the object
-NumpyAllocator g_numpyAllocator;
+NumpyAllocator *g_numpyAllocator;
```
vrabaud, line 150: this is just for speed right ?

cv_bridge/src/module.cpp
```diff
@@ -155,7 +157,7 @@ int convert_to_CvMat2(const PyObject* o, cv::Mat& m)
     if(!o || o == Py_None)
     {
         if( !m.data )
-            m.allocator = &g_numpyAllocator;
+            m.allocator = g_numpyAllocator;
```
vrabaud, line 160: where is the pointer g_numpyAllocator being initialized ?

cv_bridge/src/module.cpp
edgarriba, line 232: in the new Mat interface doesn't exist refcount attribute anymore

cv_bridge/src/module.cpp
edgarriba, line 150: i think it's also to satisfy m.allocator

//! custom allocator
MatAllocator\* allocator;
//! and the standard allocator
static MatAllocator\* getStdAllocator();

vrabaud:
ok. You are proposing changes that do not compile under OpenCV 2.4.*. Please include a pre-processor #if to compare the OpenCV versions and use the right code, thx.

edgarriba:
the changes of this repo are not mine. I have a fork with the versions comparissons but i'm not sure how to pull request from a fork of a fork.

https://github.com/edgarriba/vision_opencv

vrabaud:
should be the same really. But your fork also need to deal with the refcount and g_numpyAllocator ?

vrabaud:
hey, I am at ROScon right now. I released OpenCV3 as a deb and I am planning to test your fixes with it !

edgarriba:
nice!

PR #46, opened 2014-09-06
Changes for OpenCV 3

edgarriba:


cv_bridge/src/cv_bridge.cpp
```diff
@@ -42,6 +42,8 @@
 
 #include <cv_bridge/cv_bridge.h>
 
+#include <opencv2/imgproc/types_c.h>
```
vrabaud, line 45: If I am not mistaken, opencv2 headers will disappear in OpenCV 3.0 no ?

edgarriba, line 45: I don't think so.

http://code.opencv.org/projects/opencv/wiki/ChangeLog

In OpenCV 3 we write
#include "opencv2/core.hpp"
instead of
#include "opencv2/core/core.hpp"

PR #50, opened 2014-11-13
Python image_geometry typos and updates for cv2

dimatura:
Hi, I'm using the Python `image_geometry` module and have noticed a couple of problems. `PinholeCameraModel` is lacking a `tfFrame` method, which breaks `StereoCameraModel.tfFrame()`. The `PinholeCamera.rectifyImage` method does not currently work, for two reasons. First, `np` is used instead of `numpy`. Second, the arguments seem to assume the old `cv` API, and fail with `cv2`. This PR fixes these problems.


vrabaud:
thx ! those were big problems.

PR #52, opened 2014-12-02
Updated Q matrix to account for cameras with different Fx and Fy values

carnegieroboticsllc:
There current conversions to pointclouds use a simplified Q matrix which assumes Fx = Fy. This updates the Q matrix to account for cameras with different Fx and Fy values.


vrabaud:
thx ! that seems very valuable.

PR #53, opened 2014-12-04
Update stereo_camera_model.cpp

carnegieroboticsllc:
Correct slight error in the Q matrix derivation


vrabaud:
how about those checks ? 

```
assert( left_.fx() == right_.fx() );
```

Are they still relevant ?

carnegieroboticsllc:
I think that is still relavent. Confirming that the user has a proper calibration for their stereo camera would not hurt.

PR #55, opened 2014-12-15
Remove Python for Android

garyservin:


vrabaud:
nice ! more platforms :)

vrabaud:
hi, do you want me to release a new version of vision_opencv just for that change ?

garyservin:
Hi! I think there's no need for a new version just for this.

PR #58, opened 2015-01-05
fix error: ISO C++ forbids initialization of member for gcc 4.6

k-okada:


vrabaud:
Right, I had seen that but I forgot some old gcc's were still on the farm. Thx !

PR #65, opened 2015-07-13
fix image output of fback_flow

aginika:
Now this will show the result of flow calculation instead of frame image.
fback_flow

### DIFF
--- a/opencv_apps/src/nodelet/fback_flow_nodelet.cpp
+++ b/opencv_apps/src/nodelet/fback_flow_nodelet.cpp
@@ -166,7 +166,7 @@ class FBackFlowNodelet : public nodelet::Nodelet
 
 
       // Publish the image.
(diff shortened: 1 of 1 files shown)


edgarriba:
Hi,

I've been check-in your code and the proper way to switch from opencv version 2/3 is in code with the variable CV_VERSION_EPOCH instead of adding a definition in the cmake. Because if someone take your code without the cmake, it will crash.

For instance:

#ifndef CV_VERSION_EPOCH
 // opencv2 code
#else
  // opencv3 code
#endif

aginika:
thanks for comment, but my change is just to publish image overray-ed by optical flow, instead of original raw image, So it is not related to opencv 2/3 issue.

vrabaud:
thx ! The OpenCV 2/3 issues will be fixed in an upcoming commit

vrabaud:
done in c2f7d81a20426988bec7042f971356340fa0a9a1

PR #71, opened 2015-09-09
Add format enum for easy use and choose format.

talregev:


cv_bridge/include/cv_bridge/cv_bridge.h
```diff
@@ -55,6 +55,16 @@ class CvImage;
...
+//from: http://docs.opencv.org/modules/highgui/doc/reading_and_writing_images_and_video.html#Mat imread(const string& filename, int flags)
+typedef enum {
+	BMP, DIP,
+	JPG, JPEG, JPE,
+	JP2,
+	PNG,
+	PBM, PGM, PPM,
+	SR, RAS,
+	TIFF, TIF,
+} Foramt;
```
vrabaud: please name it `Format`

talregev: It already have name "Format"

vrabaud: I read `Foramt`. There's a typo :)

talregev: O sorry, you are right :)

talregev: @vrabaud  change and squash ;)

cv_bridge/include/cv_bridge/cv_bridge.h
```diff
@@ -55,6 +55,16 @@ class CvImage;
...
 
+//from: http://docs.opencv.org/modules/highgui/doc/reading_and_writing_images_and_video.html#Mat imread(const string& filename, int flags)
+typedef enum {
+	BMP, DIP,
+	JPG, JPEG, JPE,
+	JP2,
+	PNG,
+	PBM, PGM, PPM,
+	SR, RAS,
+	TIFF, TIF,
```
vrabaud, line 66: TIF should be before TIFF in alphabetical order

vrabaud, line 66: does not matter actually, plus PNG already was before. That's fine no worries.

cv_bridge/src/cv_bridge.cpp
```diff
@@ -409,14 +409,49 @@ cv::Mat matFromImage(const sensor_msgs::CompressedImage& source)
...
+		case RAS:
+			return "ras";
+		case SR:
+			return "sr";
+		case TIF:
+			return "tif";
+		case TIFF:
+			return "tiff";
+	}
+	return "";
```
vrabaud: should be throw an error here maybe ?

talregev: I don't think so, because it enum, I cover all the possibilities , user cannot enter a wrong format.

talregev: If you insist I will add a a throw there.
Any more feedback?

vrabaud: It's more that the behavior is undefined right ? What if I put an int that is bigger than the biggest enum in Format ? You can also have a warning message that says the format is not supported and default to JPG.

talregev:
ok

cv_bridge/src/cv_bridge.cpp
talregev: ok. add a throw.

talregev: How I write a warning message?
Write to std error?

talregev:
thanks :)

PR #75, opened 2015-09-29
Added conversion support for kinect depth images.

carlosmccosta:
Added sensor_msgs::Image conversion to cv::Mat from rqt_image_view in order to be able to create videos from kinect color and depth images
(cv_bridge
currently doesn't support 16UC1 image encoding).

Related PR:
https://github.com/ros-perception/image_pipeline/pull/147

Code adapted from:
https://github.com/ros-visualization/rqt_common_plugins/blob/groovy-devel/rqt_image_view/src/rqt_image_view/image_view.cpp

### DIFF
--- a/cv_bridge/include/cv_bridge/cv_bridge.h
+++ b/cv_bridge/include/cv_bridge/cv_bridge.h
@@ -38,8 +38,11 @@
 
 #include <sensor_msgs/Image.h>
 #include <sensor_msgs/CompressedImage.h>
+#include <sensor_msgs/image_encodings.h>
 #include <ros/static_assert.h>
 #include <opencv2/core/core.hpp>
+#include <opencv2/imgproc/imgproc.hpp>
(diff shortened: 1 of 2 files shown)


vrabaud:
I have not merged that as there are a few related issues. I'd rather have a cvtColorForDisplay function that would use the same rules from rqt_image_view and that would have a few parameters, of which range would be (but not necessarily for depth). I just need a bit of time but if you're willing to do it, thx !

carlosmccosta:
Thanks for the feedback :)
I refactored the code to use cv_bridge::toCvShare instead of cv_bridge::toCvCopy (for efficiency reasons) and changed the function to the requested name.
I also changed the docs and function parameters in order to not give the impression that the parameters values are for depth images only.

Have a nice day :)

vrabaud:
I was out for a month, sorry. I am not entirely satisfied with the result but I'll make another PR for discussion. Thx !

PR #76, opened 2015-10-05
Add opencv_apps and opencv_tests to vision_opencv meta package

garaemon:


vrabaud:
I would be against opencv_tests : they are really tests, and are undocumented. Most functionalities are probably in opencv_apps too.
Sure for opencv_apps.

garaemon:
So, Do I need to remove opencv_tests?

vrabaud:
yes please, do remove it for now

Those tests actually do not make sense anymore: they could be part of opencv_apps. There are not real tests, just basic apps.

garaemon:
updated

garaemon:
:bow:

PR #77, opened 2015-10-05
Cleanup opencv_apps CMakeLists.txt by using link_libraries macro

garaemon:
Remove many `target_link_libraries` and use one `link_libraries`


vrabaud:
link_libraries is deprecated: https://cmake.org/cmake/help/v3.0/command/link_libraries.html?highlight=link_libraries

vrabaud:
any good reason why you wanted to use it ? Or just clarity ?

garaemon:
> just clarity ?

Yes

garaemon:
> link_libraries is deprecated: https://cmake.org/cmake/help/v3.0/command/link_libraries.html?highlight=link_libraries

oh...

PR #81, opened 2015-11-15
add more possiblities to cvtColorForDisplay

vrabaud:


vrabaud:
Hi all, sorry, I was off the grid for a month. What do you guys think of the following to final get homogeneous color conversions for display purposes ?
@carlosmccosta , that is in response to https://github.com/ros-perception/vision_opencv/pull/75 and https://github.com/ros-perception/image_pipeline/pull/147
@laurent-george, it is in response to https://github.com/ros-perception/image_pipeline/pull/150 (even though your Python contribution will still be valid)
@wkentaro and @k-okada, it is in response to https://github.com/ros-perception/image_pipeline/pull/157

Your 3 pull requests, could then simply be replaced by:
https://github.com/vrabaud/image_pipeline/commit/46667b4a0255fe15d9c17da0b1734a5e411dd0b9

I'd rather have the logic for converting to displayable data in one place for use by others (like rqt_image_view for example that has its own logic: https://github.com/ros-visualization/rqt_common_plugins/blob/groovy-devel/rqt_image_view/src/rqt_image_view/image_view.cpp#L285 )

Comments ?

wkentaro:
I like the idea putting this kind of function in a more upstream package, and each package (rqt_image_view, image_view and maybe rviz image_view) depends on that. 
I wonder why this function is not in OpenCV. What do you think about this @vrabaud?
(you seems to contributing to OpenCV frequently)

vrabaud:
hmm. OpenCV already has its own logic, cf http://docs.opencv.org/2.4/modules/highgui/doc/user_interface.html#imshow . This function is so small, easy to write and dependent on your convention that I don't think it fits in OpenCV. But vision_opencv is good enough :)

And good catch on RViz ! It indeed has a lot of custom code: https://github.com/ros-visualization/rviz/blob/indigo-devel/src/rviz/image/ros_image_texture.cpp#L198 . And it already depends on image_geometry, hence OpenCV. Adding a small dependency on cv_bridge should not be a problem. It even also has itw own image_view: https://github.com/ros-visualization/rviz/blob/indigo-devel/src/image_view/CMakeLists.txt

wkentaro:
I see.
So do you plan to replace existing converting code with this function like below?
1. Merge this PR.
2. Replace image_view's converting code
3. Adopt good converting logics from rqt_image_view
4. Replace rqt_image_view's converting code
5. Adopt good converting logics from rviz image_view plugin
6. Replace rviz image_view plugin's converting code

vrabaud:
@wkentaro : ideally it would be more:
1. Adopt good converting logics from rqt_image_view and rviz image_view plugin
2. Merge this PR.
3. Replace image_view's converting code
4. Replace rqt_image_view's converting code
5. Replace rviz image_view plugin's converting code

Now, there might be the following problems:
- rqt_image_view would need to create a new branch as it would depend on an Indigo cv_bridge (and it can now work with groovy).
- RViz would have a new dependency: cv_bridge (which is lightweight considering that OpenCV is already a dependency).

wkentaro:
@vrabaud I see.

So 1 to 3 can be done without any problem.
I agree with your plan to adapt this function with existing software.

PR #82, opened 2015-11-16
[image_geometry]add a condition if D=None

YuOhara:
NULL D of pinhole_camera-parameter is supported in OpenCV.
http://docs.opencv.org/2.4/modules/calib3d/doc/camera_calibration_and_3d_reconstruction.html#projectpoints
Can this condition be added? without this, Segmentation fault may occur if D is null.


vrabaud:
totally makes sense. For future C++< please use the empty() member function.
Thx !

Issue #83, opened 2015-11-18
Wrong conversion from MONO16 to BGR8


stfuchs:
Noticed that when I was trying to visualize some MONO16 images with rqt_image_view.
After some digging I found that rqt_image_view first tries to convert the input image to BGR8 using the cv_bridge functions.

There is a bug in the ordering when trying to perform color conversion with images of different depths. I've prepared a patch. Will send a PR.

vrabaud:
Hi though your PR was ok for your case, it would not work for the reverse. Plus, there was a big bug in the code (SAME_FORMAT could be called twice). I fixed the tests and code. Thx for reporting !

stfuchs:
Cool, thanks for fixing.

PR #84, opened 2015-11-18
fix depth conversion

stfuchs:
When trying to convert from MONO16 to BGR8 the order should be:

MONO16 -> MONO8
MONO8 -> BGR8

Currently we first try to call cvtColor with images of different depths and than try to fix the depth. This doesn't work.

see #83


vrabaud:
see #83 for a more generic fix. Thx for hinting at the error !

PR #88, opened 2015-12-02
[opencv_apps] fix bug for segment_objects_nodelet.cpp

knorth55:
error for OpenCV 2

``` sh
$ roslaunch usb_cam usb_cam-test.launch 
$ rosrun opencv_apps segment_objects image:=/usb_cam/image_raw
OpenCV Error: Bad argument (No parameter 'noiseSigma' is found) in set, file /build/buildd/opencv-2.4.8+dfsg1/modules/core/src/algorithm.cpp, line 638
```


k-okada:
thank you, here is the regression -> https://github.com/ros-perception/vision_opencv/commit/89a933aef7c11acdb75a17c46bfcb60389b25baa#diff-c6a2c6a217a47520775da38df1817c51L79

PR #92, opened 2015-12-12
Convert label to BGR image to display

### LINKED ISSUE #90: [cv_bridge] Support 32SC1 to visualize in cvtColorForDisplay
I think 32SC1 is usually used for label image, and label image should be visualize with converting it to RGB image.
like [skimage.color.label2rgb](http://scikit-image.org/docs/dev/api/skimage.color.html#label2rgb)

wkentaro:
Closes #90

### DIFF
--- a/cv_bridge/include/cv_bridge/cv_bridge.h
+++ b/cv_bridge/include/cv_bridge/cv_bridge.h
@@ -264,6 +264,8 @@ CvImagePtr cvtColor(const CvImageConstPtr& source,
  * isMono return true. It must also be 8 bit in depth
  * - if the input encoding is an OpenCV format (e.g. 8UC1), and if we have 1,3 or 4 channels, it is
  * respectively converted to mono, BGR or BGRA.
+ * - if the input encoding is 32SC1, this estimate that image as label image and will convert it as
+ * bgr image with different colors for each label.
  *
  * \param source   A shared_ptr to a sensor_msgs::Image message
  * \param encoding Either an encoding string that returns true in sensor_msgs::image_encodings::isColor
--- a/cv_bridge/include/cv_bridge/rgb_colors.h
+++ b/cv_bridge/include/cv_bridge/rgb_colors.h
@@ -1,0 +1,211 @@
+// -*- mode: c++ -*-
+/*********************************************************************
+ * Software License Agreement (BSD License)
+ *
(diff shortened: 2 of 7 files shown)


k-okada:
LGTM

wkentaro:
@vrabaud Could you please review this?

vrabaud:
ok, that needs doc in the cvtForDisplay header.
Where did you get those colors from ? I hope it's not protected by a trademark or something.
Very nice to have otherwise.

wkentaro:
Thank you for pointing out about color licenses.
The color definition is copied from scikit-image. https://github.com/scikit-image/scikit-image/blob/master/skimage/color/rgb_colors.py
The license is Modified BSD, and the color definition seems not protected though.

vrabaud:
why not using colormap by the way ? http://docs.opencv.org/2.4/modules/contrib/doc/facerec/colormaps.html
It would simplify the code and license.

wkentaro:
I didn't know that feature in opencv, thank you for sharing with me.

I tried that function and I think the problem of that function is the color changes gradually so different label looks same label some times.

**with rgbColors**
image

**with applyColorMap**
image

wkentaro:
it would be like https://github.com/ros-perception/vision_opencv/pull/95

vrabaud:
well, you have to think about the cases you want to solve:
- do I know the number of indices before hand ?
- can this number be predefined ? (like you did in your PR).
- do I want to change the map ? (which makes sense for depth or cluster indices)
  You can also compute those colors in a cube at runtime to make sure they are as far from each other as possible.

wkentaro:
- do I know the number of indices before hand ? (-> no, but the unique pixel values in label image is the number in most cases)
- can this number be predefined ? (like you did in your PR). (-> In my PR, it can differentiate up to 146 clusters, in most case it's enough I think)
- do I want to change the map ? (which makes sense for depth or cluster indices) You can also compute those colors in a cube at runtime to make sure they are as far from each other as possible. (-> For depth image, yes, but for label image I don't need to change the map)

vrabaud:
ok, let's merge that in. Please:
- rebase
- add a comment to say where you got the colors from and add the project in the copyright of the BSD license.
  Thx

wkentaro:
I'm done.

wkentaro:
Closes #95

PR #100, opened 2016-01-02
set image encoding to bgr8

k-okada:
to support both color and mono input image

### DIFF
--- a/opencv_apps/src/nodelet/simple_flow_nodelet.cpp
+++ b/opencv_apps/src/nodelet/simple_flow_nodelet.cpp
@@ -113,7 +113,15 @@ class SimpleFlowNodelet : public nodelet::Nodelet
     try
     {
       // Convert the image into something opencv can handle.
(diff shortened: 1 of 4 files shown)


opencv_apps/src/nodelet/simple_flow_nodelet.cpp
```diff
@@ -113,7 +113,15 @@ class SimpleFlowNodelet : public nodelet::Nodelet
     try
     {
       // Convert the image into something opencv can handle.
-      cv::Mat frame = cv_bridge::toCvShare(msg, msg->encoding)->image;
+      cv::Mat frame_src = cv_bridge::toCvShare(msg, msg->encoding)->image;
```
vrabaud, line 116: why not use an encoding of "bgr8" here ?

k-okada, line 116: because this line convert ros msg to opencv Mat, so we should set
msg->encoding.
int this node, we use frame_src as opencv Mat with input encocding and
frame for bgr8 image for output

## 

◉ Kei Okada

PR #103, opened 2016-01-02
add test programs for opencv_apps

k-okada:
This may take too long to download bag files.
- opencv_apps: add test programs, this will generate images for wiki
- src/nodelet/segment_objects_nodelet.cpp: change output image topic name from segmented to image

### DIFF
--- a/opencv_apps/package.xml
+++ b/opencv_apps/package.xml
@@ -27,6 +27,15 @@
   <run_depend>roscpp</run_depend>
   <run_depend>std_srvs</run_depend>
 
+  <test_depend>rostest</test_depend>
+  <test_depend>rosbag</test_depend>
+  <test_depend>rosservice</test_depend>
+  <test_depend>rostopic</test_depend>
+  <test_depend>procps</test_depend>
+  <test_depend>image_proc</test_depend>
+  <test_depend>image_view</test_depend>
+  <test_depend>topic_tools</test_depend>
+
   <export>
     <nodelet plugin="${prefix}/nodelet_plugins.xml"/>
   </export>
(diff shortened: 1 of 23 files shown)


vrabaud:
It seems testwatershed_segmentation_test is failing.

k-okada:
@vrabaud ok, travis passed, now ready to review.

vrabaud:
nice catch for the revert.
Why do you also need image_pipeline to be recompiled ? If a release needs to be made for image_pipeline, please let me know and I'll make one (and please remove that .travis modification (even if it will break .travis for now)).
Also, please squash your commits (it's better when every commit passes Travis). Thx ! tests are always welcome :)

k-okada:
we need to recompile image_pipeline for using image_proc in the test code,
like
https://github.com/k-okada/vision_opencv/blob/add_test/opencv_apps/test/test-contour_moments.test#L6,
another option is to not using image_proc in that launch file...

## 

◉ Kei Okada

vrabaud:
why not using image_proc as a test_depend in package.xml ?

k-okada:
because of https://github.com/ros-perception/image_pipeline/pull/172

## 

◉ Kei Okada

k-okada:
building source from image_proc is  only needed for opencv3 test.

vrabaud:
Right, totally missed that ... Thx !

opencv_apps/package.xml
```diff
@@ -27,6 +27,15 @@
   <run_depend>roscpp</run_depend>
   <run_depend>std_srvs</run_depend>
 
+  <test_depend>rostest</test_depend>
+  <test_depend>rosbag</test_depend>
+  <test_depend>rosservice</test_depend>
+  <test_depend>rostopic</test_depend>
+  <test_depend>procps</test_depend>
+  <test_depend>image_proc</test_depend>
+  <test_depend>image_view</test_depend>
```
dirk-thomas, line 36: This line introduces a circular dependency on the repository level between this repo (B) and  `image_pipeline` (A) . E.g. the package `camera_calibration`in A depends on `cv_bridge` in B and `opencv_apps` in B depends on `image_view` in A.

As a result none of the two repos can be released into a new distro without ignoring packages.

PR #105, opened 2016-01-04
cv_bridge: add OpenCV_INCLUDE_DIRS to INCLUDE_DIRS

k-okada:
when we have installed both libopencv-dev and ros-indigo-opencv3, this will export include dir which is used for compilation time of cv_bridge

### DIFF
--- a/cv_bridge/CMakeLists.txt
+++ b/cv_bridge/CMakeLists.txt
@@ -11,7 +11,7 @@ endif()
 find_package(OpenCV REQUIRED)
 
 catkin_package(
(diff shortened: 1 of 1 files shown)


trainman419:
See https://github.com/ros-perception/image_pipeline/pull/172 for comments on why this is not a good idea.

vrabaud:
This PR is invalid as INCLUDE_DIRS has to be in source:
http://docs.ros.org/groovy/api/catkin/html/dev_guide/generated_cmake_api.html#catkin-package

Issue #106, opened 2016-01-22
procps which is a test dependency is not available on OS X (or any of the BSD's)


spaghetti-:
Defined here: https://github.com/ros-perception/vision_opencv/blob/indigo/opencv_apps/package.xml#L34

Commited by @k-okada at https://github.com/ros-perception/vision_opencv/commit/99f5a4d61771426dd3a2db35473a2a9aa9105806

I couldn't find any uses for the program in the tests. What exactly is the reason for this change?

Thanks

k-okada:
humm, I'm very sorry that I can't recall why I added this, #107 is the PR
to remove this.
I have no knowledge why procps is not available on BSD machine, but you can
add dummy entry to rosdep file, for example

```
procps:
  macports: []
```

to avoid this type of error

## 

◉ Kei Okada

spaghetti-:
@k-okada no problem, just ran into this issue and I actually had a PR in progress with the same change, but looks like you've got it. Thanks for the quick change.

k-okada:
I see, I used this for `watch` program at https://github.com/ros-perception/vision_opencv/blob/indigo/opencv_apps/test/test-lk_flow.test#L17, periodically update optical flow. What is the equivalent program for OS X?

spaghetti-:
Homebrew seems to have https://gitlab.com/procps-ng/procps which is essentially the same. Maybe you can revert the merge and I'll submit a PR to rosdistro with this as required for OS X?

spaghetti-:
However, I do apologize for not doing my research first and going purely based off /proc.

trainman419:
This showed up on ROS Answers: http://answers.ros.org/question/224956/no-definition-of-procps-for-os-osx/

This does not appear to be fixed; #107 was closed without merging, and nothing was submitted to rosdep.

I think https://github.com/ros-perception/vision_opencv/blob/indigo/opencv_apps/test/test-lk_flow.test#L17 can be better done using `respawn="true"` and `respwan_delay=2` instead of relying on a linux-specific tool.

acornellier:
I am still having this issue.

ERROR: the following packages/stacks could not have their rosdep keys resolved
to system dependencies:
opencv_apps: No definition of [procps] for OS [osx]

acornellier:
I was able to fix it buy removing a line from vision_open_cv/opencv_apps/package.xml:

```
<test_depend>procps</test_depend>
```

vrabaud:
[...]

PR #107, opened 2016-01-22
remove procps from test_depend, #106

k-okada:


k-okada:
os x may have which command, better to update rosdistro package

PR #109, opened 2016-01-27
Use respawn instead of watch

trainman419:
Proposed fix for #106 . Tests pass on Ubuntu, without the use of `watch`


k-okada:
humm, I didn't think of this. nice.

## 

◉ Kei Okada

vrabaud:
will release as soon as #110 is in.

Issue #108, opened 2016-01-26
Question regarding OpenCV3, optflow.hpp and opencv-contrib dependency


bchretien:
Hi!

I'm maintaining ROS packages for Arch Linux, and a user (cc @Caopi) [reported](https://aur.archlinux.org/packages/ros-indigo-opencv-apps/) the following problem: when compiling `opencv_apps` with OpenCV3 installed,  the compilation failed for `simple_flow_nodelet.cpp` because `optflow.hpp` is missing (see the related commit: https://github.com/ros-perception/vision_opencv/commit/8ed5ff5c48b4c3d270cd8216175cf6a8441cb046). Currently, the only solution is to depend on the `opencv-contrib` package. Note that on Arch, the default OpenCV version is still OpenCV2.

The OpenCV2 version of `simple_flow_nodelet.cpp` relies on the standard `video/tracking.hpp` header, but the OpenCV3 version expects an `opencv-contrib` header. Are the `opencv-contrib` modules supposed to be standard for an OpenCV3 install? If not, that exe should be made optional with proper `opencv-contrib` checks in `CMakeLists.txt`, or the code should be modified to only depend on the standard OpenCV package.

vrabaud:
OpenCV3 is not officially supported in ROS.
Now, there is an OpenCV3 package that has been released that does not conflict with Ubuntu's OpenCV2. It has not been tested on Arch but that would be the one that should become the official one and yes, it includes opencv-contrib. How it is built is explained at the bottom of http://wiki.ros.org/action/show/vision_opencv?action=show&redirect=opencv
I am trying to get OpenCV3 as default in Kinetic and I need to create an OpenCV3 wiki page.
Any custom build of OpenCV2 or 3 is not supported.
Does that answer all your questions ?

mschlaffer:
Thanks for your answer vrabaud. I understand where you're coming from and appreciate your efforts to get OpenCV3 as default in Kinetic. 

The problem is not to get OpenCV 3 to work in Arch, there is a [package](https://aur.archlinux.org/packages/opencv-contrib/) to get it with all contrib libraries installed and it does work with ros-indigo. The rest is just a dependency issue or something that needs to be clarified in installation instructions.

[...]

PR #110, opened 2016-01-27
check if OpenCV_LIB_COMPONENTS contains opencv_contrib

k-okada:
see #108 for discussion

### DIFF
--- a/opencv_apps/CMakeLists.txt
+++ b/opencv_apps/CMakeLists.txt
@@ -4,6 +4,10 @@ project(opencv_apps)
 find_package(catkin REQUIRED COMPONENTS cv_bridge dynamic_reconfigure message_generation image_transport nodelet roscpp)
 
 find_package(OpenCV REQUIRED)
+message(STATUS "OpenCV Components: ${OpenCV_LIB_COMPONENTS}")
+if(OpenCV_VERSION VERSION_LESS "3.0" OR TARGET opencv_optflow)
+  set(OPENCV_HAVE_OPTFLOW TRUE)
+endif()
 
 # generate the dynamic_reconfigure config file
 generate_dynamic_reconfigure_options(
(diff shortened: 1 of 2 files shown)


opencv_apps/CMakeLists.txt
```diff
@@ -70,6 +70,10 @@ catkin_package()
 include_directories( ${catkin_INCLUDE_DIRS} ${OpenCV_INCLUDE_DIRS})
 
 ## Declare a cpp library
+if("${OpenCV_LIB_COMPONENTS}" MATCHES ";opencv_contrib;")
```
vrabaud: I'd rather keep it specific  `if("${OpenCV_LIB_COMPONENTS}" MATCHES "opencv_optflow")` no ?

vrabaud: Better please use `if(TARGET opencv_optflow)`

vrabaud:
You should also disable test-simple_flow.test

opencv_apps/CMakeLists.txt
k-okada: on indigo/opencv2 environment, we do not have opencv_optflow nor TARGET openecv_optflow

vrabaud: oh right, then `if(OpenCV_VERSION VERSION_LESS "3.0" OR TARGET opencv_optflow)`

trainman419:
Instead of calling the check to see if optflow exists multiple times, it's probably better to call that check once, and set a variable if optflow exists; then all of the other compilation steps that depend on optflow can use that variable.

```
if( ... however you decide to detect optflow ... )
     set(OPENCV_HAVE_OPTFLOW TRUE)
endif()

...

if(${OPENCV_HAVE_OPTFLOW})
   # build something that needs optflow
endif()

# repeat as needed for tests, etc
```

vrabaud:
agreed with @trainman419 
- disable the test too. Thx.

k-okada:
fixed, see latest commit

## 

◉ Kei Okada

opencv_apps/CMakeLists.txt
```diff
@@ -70,6 +74,10 @@ catkin_package()
 include_directories( ${catkin_INCLUDE_DIRS} ${OpenCV_INCLUDE_DIRS})
 
 ## Declare a cpp library
+if(${OPENCV_HAVE_OPTFLOW})
```
vrabaud: The proper way to check for a variable is `if(OPENCV_HAVE_OPTFLOW)`. https://cmake.org/cmake/help/v3.0/command/if.html
`If(<variable>)` is a bit like what is done above with `OpenCV_VERSION`.

k-okada: ok, fixed

## 

◉ Kei Okada

vrabaud:
Thx a bunch !

PR #111, opened 2016-02-11
Refactor opencv_apps to remove duplicated codes to handle connection of topics.

garaemon:
1. Add opencv_apps::Nodelet class to handle connection and disconnection of
   topics.
2. Update nodelets of opencv_apps to inhereit opencv_apps::Nodelet class
   to remove duplicated codes.


vrabaud:
Very nice refactoring, thx

PR #112, opened 2016-02-25
Remove eval, and other fixes

eric-wieser:
Also, extend from object, so as not to get a python 2.2-style class, and use the new-style raise statement.

There's very rarely a good reason to use eval, and this is not an exception to that rule

### DIFF
--- a/cv_bridge/python/cv_bridge/core.py
+++ b/cv_bridge/python/cv_bridge/core.py
@@ -38,7 +38,7 @@ class CvBridgeError(TypeError):
     """
     pass
 
-class CvBridge:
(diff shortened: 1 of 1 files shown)


eric-wieser:
This is actually really bad - I can craft a malicious image packet, that when received by another node, possibly on a machine I do not control, causes arbitrary code execution!

```
from sensor_msgs.msg import Image

some_publisher.publish(Image(...,
    encoding='8UC__import__("os").system("rm -rf /")'
))
```

vrabaud:
nice catches ! Very bad indeed ... Will release soonish.

Issue #113, opened 2016-02-25
failing frequency tests


vrabaud:
@k-okada , @garaemon , this would be on your side:
http://build.ros.org/job/Idev__vision_opencv__ubuntu_trusty_amd64/10/testReport/__main__/HzTest/test_hz/
I guess it makes sense as those depend on the speed of the machine. Maybe cutting the rate in half would work ?

vrabaud:
More here: http://build.ros.org/job/Jdev__vision_opencv__ubuntu_trusty_amd64/10/testReport/

vrabaud:
fixed in #116

PR #116, opened 2016-02-29
pass test on slower machines

k-okada:
see #113 for issue


vrabaud:
Thx, that should close #113

PR #119, opened 2016-03-15
geometry_msgs doesn't get used by opencv_apps, but std_msgs does.

lucasw:
I tried making a package that used opencv_apps msgs, but then the build complained about geometry_msgs, which I thought opencv_apps ought to be putting into catkin_package()- but it looks like it doesn't need it at all.

### DIFF
--- a/opencv_apps/CMakeLists.txt
+++ b/opencv_apps/CMakeLists.txt
@@ -66,10 +66,14 @@ add_message_files(
 ## Generate added messages and services with any dependencies listed here
 generate_messages(
   DEPENDENCIES
-  geometry_msgs
+  std_msgs
 )
 
-catkin_package()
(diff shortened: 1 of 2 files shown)


vrabaud:
looks justified. Would a release be needed just for that ?

lucasw:
I don't think a release is needed, it could wait until some other changes pile up.  

Is there a way to see what released packages depend on opencv_apps?  (Other than having local copy of every package source and grepping their package.xmls)  If there are some and they could eliminate the geometry_msgs dependency that would be more motivation to release earlier.

vrabaud:
Apparently none:
http://rosindex.github.io/p/opencv_apps/github-ros-perception-vision_opencv/#indigo-deps

PR #123, opened 2016-04-06
opencv_apps/CMakeLists.txt: add missing install target

k-okada:


vrabaud:
opencv_apps has moved and you have write access to it: https://github.com/ros-perception/opencv_apps

PR #126, opened 2016-04-22
[cv_bridge] Don't colorize float image as label image

wkentaro:
This is a bug and image whose encoding is other than 32SC1 should not be
colorized. (currently, depth images with 32FC1 is also colorized.)


vrabaud:
Thx !

PR #127, opened 2016-05-01
Add fullResolution getter to python PinholeCameraModel

jpanikulam:
A feature often useful in C++, but not available in Python

Also, would a PR for improving numpy support be welcome? [Here's the kludge](https://github.com/uf-mil/Sub8/blob/master/gnc/sub8_perception/sub8_vision_tools/estimation.py#L164) I've resorted to elsewhere.

Love, Jake

### DIFF
--- a/image_geometry/src/image_geometry/cameramodels.py
+++ b/image_geometry/src/image_geometry/cameramodels.py
@@ -52,6 +52,8 @@ class PinholeCameraModel:
         self.height = msg.height
         self.binning_x = max(1, msg.binning_x)
         self.binning_y = max(1, msg.binning_y)
+        self.resolution = (msg.width, msg.height)
+
         self.raw_roi = copy.copy(msg.roi)
         # ROI all zeros is considered the same as full resolution
         if (self.raw_roi.x_offset == 0 and self.raw_roi.y_offset == 0 and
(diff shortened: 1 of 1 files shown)


vrabaud:
Thx ! And definitely for Numpy support, a few issues got opened. I don't understand your commit, what do you want to improve ?

jpanikulam:
Numpy users often expect to be able to send a an array of points and get a array of answers back out.

For example, in [project3dToPixel](https://github.com/ros-perception/vision_opencv/blob/indigo/image_geometry/src/image_geometry/cameramodels.py#L109), right now you must pass a single point.

I would like to make it easy to pass an array of points to most of the functions in PinholeCameraModel. The way people use it right now would remain the same, but it would be easier to process many points at once.

Will get to work on it!

vrabaud:
ok, that totally makes sense. Please do ! And don't forget the test ;) Do we still have an interest in having this pure Python API ? Shouldn't we wrap it with Boost Python like I did with cv_bridge ?

jpanikulam:
I am a fan of the pure Python API, it's fun for the whole family!

vrabaud:
fun but requires double maintenance. What you suggest is totally valid but will have to be duplicated in C++.

jpanikulam:
I see. In the short term, improving the Python api to be fully compatible with cv2 is super easy, ~2 hrs of work. 

Writing bindings looks like a fair amount more, but the end result is no duplicated implementation. What would you prefer?

jpanikulam:
When I say "fully compatible" I mean "behave the same way", taking numpy arrays of points and returning numpy arrays of points

vrabaud:
Right, for now, please do implement the Python API but do include tests. This way, whenever we switch to bindings, we'll make sure the behavior is the same.
Wrapping is not that hard, cf what I did with cv_bridge: https://github.com/ros-perception/vision_opencv/blob/indigo/cv_bridge/src/module.cpp#L55
but probably requires more tests, so yours will be welcome. Thx !

PR #128, opened 2016-05-05
Fix compilation of cv_bridge with opencv3 and python3.

de-vri-es:
This PR fixed compilation of cv_bridge with opencv3 and python3.

The Int type from python2 no longer exists in python3, only the Long type remains. The suggested fix here [1] is to alias `PyInt_*` to `PyLong_*`, so that is what I did. Maybe it would be fine to use `PyLong` in python2 too, but I don't know if that is the case, so I went for this approach instead.

[1] https://wiki.python.org/moin/PortingExtensionModulesToPy3k

### DIFF
--- a/cv_bridge/src/module_opencv3.cpp
+++ b/cv_bridge/src/module_opencv3.cpp
@@ -6,6 +6,7 @@
 
 #include "opencv2/opencv_modules.hpp"
 
+#include "pycompat.hpp"
 
 static PyObject* opencv_error = 0;
 
--- a/cv_bridge/src/pycompat.hpp
+++ b/cv_bridge/src/pycompat.hpp
@@ -1,0 +1,70 @@
+/*M///////////////////////////////////////////////////////////////////////////////////////
+//
+//  IMPORTANT: READ BEFORE DOWNLOADING, COPYING, INSTALLING OR USING.
+//
+//  By downloading, copying, installing or using the software you agree to this license.
+//  If you do not agree to this license, do not download, install,
+//  copy or use the software.
+//
+//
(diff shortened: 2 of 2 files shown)


vrabaud:
Hi, thx that is very important. I have not checked the compatibility with Python3 in a while.
Just curious, which platform do you have ? How are the other low-level image_pipeline/image_common doing with Python 3 ?
Can you please change your PR to actually include that file https://github.com/Itseez/opencv/blob/master/modules/python/src2/pycompat.hpp
(just copy it over). I think it is better to stay close to upstream. Thx a bunch !

de-vri-es:
Good point, updated.

de-vri-es:
By the way, I'm running Archlinux, trying to compile a basic Kinetic system with python3. I haven't actually run any python code yet, I'm currently still busy patching things to compile.

Whenever I encounter an incompatibility I try to patch it and send a PR, so far I also had to patch `catkin_pkg`, `tf` and `tf2_py`.  I'm currently only compiling a subset of `ros-kinetic-desktop` though, so I will probably miss some problems.

vrabaud:
Thx for your work, that is very valuable.
Two advice:
- make sure you also build and run the test
- work on Kinetic where the current development is happening. It is also an LTS so it is becoming the new big ROS release. I forward ported your fix to Kinetic.

de-vri-es:
Thanks for porting to Kinetic.

To avoid others doing the same, maybe you could set the default branch to kinetic? I just did a clone and started working from there.

vrabaud:
indeed ! I had totally forgotten about that, thx ! Done.

PR #131, opened 2016-05-13
add cv2_to_cmprsimgmsg

talregev:
add cv2_to_cmprsimgmsg
- convert from cv2 image to compressed image ros msg.

### DIFF
--- a/cv_bridge/python/cv_bridge/core.py
+++ b/cv_bridge/python/cv_bridge/core.py
@@ -1,6 +1,7 @@
 # Software License Agreement (BSD License)
 #
 # Copyright (c) 2011, Willow Garage, Inc.
+# Copyright (c) 2016, Tal Regev.
 # All rights reserved.
 #
 # Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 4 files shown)


talregev:
@vrabaud, I am not sure why the build is fail. I check the python code, and it work correctly.

talregev:
@vrabaud can you check and approve?

vrabaud:
The error in Travis is explicit: you are using rospy.Time.now() which is not instantiated.
You should not use that but have it be a parameter of the functions.
Also, cmprsimgmsg_to_cv2 is a pretty bad naming: I read it as compares image message to cv2.
compressed_img_msg_to_cv2 maybe ?

talregev:
ok. Thanks for your feedback! :+1:

talregev:
@vrabaud please tell me what you think.

talregev:
@vrabaud Thanks for your help.
I squash the commits, can you check and approve?

Thanks, 
Tal.

vrabaud:
To follow on #132

PR #132, opened 2016-05-19
Support compressed Images messages in python for kinetic

talregev:
- Add cv2_to_comprssed_imgmsg: Convert from cv2 image to compressed image ros msg.
- Add comprssed_imgmsg_to_cv2:   Convert the compress message to a new image.
- Add compressed image tests.
- Add time to msgs (compressed and regular).

### DIFF
--- a/cv_bridge/python/cv_bridge/core.py
+++ b/cv_bridge/python/cv_bridge/core.py
@@ -1,6 +1,7 @@
 # Software License Agreement (BSD License)
 #
 # Copyright (c) 2011, Willow Garage, Inc.
+# Copyright (c) 2016, Tal Regev.
 # All rights reserved.
 #
 # Redistribution and use in source and binary forms, with or without
@@ -32,11 +33,13 @@
 
 import sensor_msgs.msg
 
+
 class CvBridgeError(TypeError):
     """
     This is the error raised by :class:`cv_bridge.CvBridge` methods when they fail.
     """
     pass
+
 
 class CvBridge(object):
     """
@@ -53,6 +56,8 @@ class CvBridge(object):
            >>> im = np.ndarray(shape=(480, 640, n_channels), dtype=dtype)
            >>> msg = br.cv2_to_imgmsg(im)  # Convert the image to a message
            >>> im2 = br.imgmsg_to_cv2(msg) # Convert the message to a new image
+           >>> cmprsmsg = br.cv2_to_compressed_imgmsg(im)  # Convert the image to a compress message
+           >>> im22 = br.compressed_imgmsg_to_cv2(msg) # Convert the compress message to a new image
            >>> cv2.imwrite("this_was_a_message_briefly.png", im2)
 
     """
@@ -61,14 +66,14 @@ class CvBridge(object):
         import cv2
         self.cvtype_to_name = {}
 
-        for t in ["8U", "8S", "16U", "16S", "32S", "32F", "64F" ]:
-            for c in [1,2,3,4]:
+        for t in ["8U", "8S", "16U", "16S", "32S", "32F", "64F"]:
+            for c in [1, 2, 3, 4]:
                 nm = "%sC%d" % (t, c)
                 self.cvtype_to_name[getattr(cv2, "CV_%s" % nm)] = nm
 
-        self.numpy_type_to_cvtype = {'uint8':'8U', 'int8':'8S', 'uint16':'16U',
-                                        'int16':'16S', 'int32':'32S', 'float32':'32F',
-                                        'float64':'64F'}
+        self.numpy_type_to_cvtype = {'uint8': '8U', 'int8': '8S', 'uint16': '16U',
+                                        'int16': '16S', 'int32': '32S', 'float32': '32F',
+                                        'float64': '64F'}
         self.numpy_type_to_cvtype.update(dict((v, k) for (k, v) in self.numpy_type_to_cvtype.items()))
 
     def dtype_with_channels_to_cvtype2(self, dtype, n_channels):
@@ -89,6 +94,46 @@ class CvBridge(object):
 
     def encoding_to_dtype_with_channels(self, encoding):
         return self.cvtype2_to_dtype_with_channels(self.encoding_to_cvtype2(encoding))
+
+    def compressed_imgmsg_to_cv2(self, cmprs_img_msg, desired_encoding = "passthrough"):
+        """
+        Convert a sensor_msgs::CompressedImage message to an OpenCV :cpp:type:`cv::Mat`.
+
+        :param cmprs_img_msg:   A :cpp:type:`sensor_msgs::CompressedImage` message
+        :param desired_encoding:  The encoding of the image data, one of the following strings:
+
+           * ``"passthrough"``
+           * one of the standard strings in sensor_msgs/image_encodings.h
+
+        :rtype: :cpp:type:`cv::Mat`
+        :raises CvBridgeError: when conversion is not possible.
+
+        If desired_encoding is ``"passthrough"``, then the returned image has the same format as img_msg.
+        Otherwise desired_encoding must be one of the standard image encodings
+
+        This function returns an OpenCV :cpp:type:`cv::Mat` message on success, or raises :exc:`cv_bridge.CvBridgeError` on failure.
+
+        If the image only has one channel, the shape has size 2 (width and height)
+        """
+        import cv2
+        import numpy as np
+
+        str_msg = cmprs_img_msg.data
+        buf = np.ndarray(shape=(1, len(str_msg)),
+                          dtype=np.uint8, buffer=cmprs_img_msg.data)
+        im = cv2.imdecode(buf, cv2.IMREAD_ANYCOLOR)
+
+        if desired_encoding == "passthrough":
+            return im
+
+        from cv_bridge.boost.cv_bridge_boost import cvtColor2
+
+        try:
+            res = cvtColor2(im, "bgr8", desired_encoding)
+        except RuntimeError as e:
+            raise CvBridgeError(e)
+
+        return res
 
     def imgmsg_to_cv2(self, img_msg, desired_encoding = "passthrough"):
         """
@@ -132,6 +177,43 @@ class CvBridge(object):
 
         return res
 
+    def cv2_to_compressed_imgmsg(self, cvim, dst_format = "jpg"):
+        """
+        Convert an OpenCV :cpp:type:`cv::Mat` type to a ROS sensor_msgs::CompressedImage message.
+
+        :param cvim:      An OpenCV :cpp:type:`cv::Mat`
+        :param dst_format:  The format of the image data, one of the following strings:
+
+           * from http://docs.opencv.org/2.4/modules/highgui/doc/reading_and_writing_images_and_video.html
+           * from http://docs.opencv.org/2.4/modules/highgui/doc/reading_and_writing_images_and_video.html#Mat imread(const string& filename, int flags)
+           * bmp, dib
+           * jpeg, jpg, jpe
+           * jp2
+           * png
+           * pbm, pgm, ppm
+           * sr, ras
+           * tiff, tif
+
+        :rtype:           A sensor_msgs.msg.CompressedImage message
+        :raises CvBridgeError: when the ``cvim`` has a type that is incompatible with ``format``
+
+
+        This function returns a sensor_msgs::Image message on success, or raises :exc:`cv_bridge.CvBridgeError` on failure.
+        """
+        import cv2
+        import numpy as np
+        if not isinstance(cvim, (np.ndarray, np.generic)):
+            raise TypeError('Your input type is not a numpy array')
+        cmprs_img_msg = sensor_msgs.msg.CompressedImage()
+        cmprs_img_msg.format = dst_format
+        ext_format = '.' + dst_format
+        try:
+            cmprs_img_msg.data = np.array(cv2.imencode(ext_format, cvim)[1]).tostring()
+        except RuntimeError as e:
+            raise CvBridgeError(e)
+
+        return cmprs_img_msg
+
     def cv2_to_imgmsg(self, cvim, encoding = "passthrough"):
         """
         Convert an OpenCV :cpp:type:`cv::Mat` type to a ROS sensor_msgs::Image message.
@@ -152,7 +234,7 @@ class CvBridge(object):
         """
         import cv2
         import numpy as np
-        if not isinstance(cvim, (np.ndarray, np.generic) ):
+        if not isinstance(cvim, (np.ndarray, np.generic)):
             raise TypeError('Your input type is not a numpy array')
         img_msg = sensor_msgs.msg.Image()
         img_msg.height = cvim.shape[0]
@@ -170,4 +252,5 @@ class CvBridge(object):
                 raise CvBridgeError("encoding specified as %s, but image has incompatible type %s" % (encoding, cv_type))
         img_msg.data = cvim.tostring()
         img_msg.step = len(img_msg.data) / img_msg.height
+
         return img_msg
--- a/cv_bridge/test/enumerants.py
+++ b/cv_bridge/test/enumerants.py
@@ -6,6 +6,7 @@ import sensor_msgs.msg
 
 from cv_bridge import CvBridge, CvBridgeError
 
+
 class TestEnumerants(unittest.TestCase):
 
     def test_enumerants_cv2(self):
@@ -31,7 +32,7 @@ class TestEnumerants(unittest.TestCase):
     def test_encode_decode_cv2(self):
         import cv2
         import numpy as np
-        fmts = [ cv2.CV_8U, cv2.CV_8S, cv2.CV_16U, cv2.CV_16S, cv2.CV_32S, cv2.CV_32F, cv2.CV_64F ]
+        fmts = [cv2.CV_8U, cv2.CV_8S, cv2.CV_16U, cv2.CV_16S, cv2.CV_32S, cv2.CV_32F, cv2.CV_64F]
 
         cvb_en = CvBridge()
         cvb_de = CvBridge()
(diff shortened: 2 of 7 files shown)


cv_bridge/python/cv_bridge/core.py
```diff
@@ -90,6 +93,47 @@ def encoding_to_cvtype2(self, encoding):
...
+
+        This function returns an OpenCV :cpp:type:`cv::Mat` message on success, or raises :exc:`cv_bridge.CvBridgeError` on failure.
+
+        If the image only has one channel, the shape has size 2 (width and height)
+        """
+        import cv2
+        import numpy as np
+
+        str_msg = cmprs_img_msg.data
+        dtype, n_channels = self.cvtype2_to_dtype_with_channels(cv2.CV_8UC1)
```
vrabaud: why cv2.CV_8UC1 ? That is a bit confusing, you can use the proper numpy dtype in the line below.

cv_bridge/python/cv_bridge/core.py
```diff
@@ -132,6 +176,47 @@ def imgmsg_to_cv2(self, img_msg, desired_encoding = "passthrough"):
...
+           * sr, ras
+           * tiff, tif
+
+        :rtype:           A sensor_msgs.msg.CompressedImage message
+        :raises CvBridgeError: when the ``cvim`` has a type that is incompatible with ``format``
+
+
+        This function returns a sensor_msgs::Image message on success, or raises :exc:`cv_bridge.CvBridgeError` on failure.
+        """
+        import cv2, rospy, time
```
vrabaud: the bad thing about that is that it adds a dependency on rospy. I would actually not fill the time member (like done in the non-compressed version) and let the user do it by hand.

opencv_tests/nodes/broadcast_compressed.py
```diff
@@ -0,0 +1,72 @@
...
+import sys
+import time
+import math
+import rospy
+import cv2
+
+import sensor_msgs.msg
+from cv_bridge import CvBridge
+
+class source:
```
vrabaud: should be Source

opencv_tests/nodes/broadcast_compressed.py
```diff
@@ -0,0 +1,72 @@
...
+    s.spin()
+    rospy.spin()
+    outcome = 'test completed'
+  except KeyboardInterrupt:
+    print "shutting down"
+    outcome = 'keyboard interrupt'
+  rospy.core.signal_shutdown(outcome)
+
+if __name__ == '__main__':
+  main(sys.argv)
```
vrabaud: a small description of what it does would be nice.

vrabaud:
ok, so overall pretty good. For the nodes, how about making the old nodes work with both compressed and uncompressed ? That would remove some copy/paste code.
Also, it needs a test. I would suggest to add compression tests to: https://github.com/ros-perception/vision_opencv/blob/kinetic/cv_bridge/test/enumerants.py

cv_bridge/python/cv_bridge/core.py
talregev: I don't know what is the proper numpy dtype for cv2.CV_8UC1.
This code came from c++:
https://github.com/ros-perception/vision_opencv/blob/kinetic/cv_bridge/src/cv_bridge.cpp#L431
When you find example in the internet how to convert jpg data to image.
For me it more understandable like this.

opencv_tests/nodes/broadcast_compressed.py
talregev: ok. we need to change also for all tests. I copy the tests from regular tests.

## 

merge regular tests and compressed together. change to Source as well

opencv_tests/nodes/broadcast_compressed.py
talregev: ok. need to add description to regular test as well:
opencv_tests/nodes/broadcast.py

## 

merge regular tests and compressed together. add  a small description as well.

cv_bridge/python/cv_bridge/core.py
talregev: Why is a bad thing to adds a dependency on rospy?
For me as a user, I expect from  using this methods (the compressed and the regular),
that they fully add all the properties with correct data, including time.
If user want to change time, he can do it after using this method.
When I use the first time in the regular method:  cv2_to_imgmsg
I surprised that it didn't fill in the time. and i need to fill it manually.  
I add time also to regular method. 
But if you still not agree with me, I will change it, without time.

talregev:
> For the nodes, how about making the old nodes work with both compressed and uncompressed ? That would remove some copy/paste code.

I think on this in the beginning, and i ask my self what if some one change bad thing in the regular or the compressed image implantation. It good we have separate tests to check which functionality it not working. But you right we need to avoid copy/past thing. so maybe we can share identical code for both tests (regular and compressed). you have idea how to do it?

## 

merge regular tests and compressed together. 

> Also, it needs a test. I would suggest to add compression tests to: https://github.com/ros-perception/vision_opencv/blob/kinetic/cv_bridge/test/enumerants.py

ok

## 

Add test in enumerants.py

talregev:
@vrabaud 
All the changes i made in indigo branch, because in that branch there is automatic tests.
When you can, please respond at my comments, then we can proceed.

talregev:
@vrabaud I also make the changes here.
Can you check?

cv_bridge/python/cv_bridge/core.py
vrabaud: Just:

```
buf = np.ndarray(shape=(1, len(str_msg)), dtype=np.uint8, buffer=cmprs_img_msg.data)
```

cv_bridge/python/cv_bridge/core.py
vrabaud: rospy is a huge dependency, and just to convert a Python time.
Having now as a default does not make sense: now() is different from the time at which you acquired the image so it seems like a dangerous default.
The API could take a time as an argument (that would default to None and therefore (0,0)) but I would suggest this as a later PR as it would need to be added to the non-compressed API too.

opencv_tests/nodes/rosfacedetect.py
```diff
@@ -33,11 +34,13 @@
 
 if __name__ == '__main__':
 
+	#TODO add this file in the repository and make it relative to this python script. (not all people will run this from linux)
```
vrabaud: indentation please use spaces and not tabs.

opencv_tests/nodes/rosfacedetect.py
```diff
@@ -33,11 +34,13 @@
...
     haarfile = '/usr/share/opencv/haarcascades/haarcascade_frontalface_alt.xml'
 
     parser = OptionParser(usage = "usage: %prog [options]")
     parser.add_option("-c", "--cascade", action="store", dest="cascade", type="str", help="Haar cascade file, default %default", default = haarfile)
     parser.add_option("-t", "--topic", action="store", dest="topic", type="str", help="Topic to find a face on, default %default", default = '/camera/rgb/image_raw')
+	parser.add_option("-ct", "--ctopic", action="store", dest="ctopic", type="str", help="Compressed topic to find a face on, default %default", default = '/camera/rgb/image/compressed')
```
vrabaud: indentation

opencv_tests/nodes/rosfacedetect.py
```diff
@@ -70,6 +73,34 @@ def detect_and_draw(imgmsg):
...
+                    pt2 = (int((x + w) * image_scale), int((y + h) * image_scale))
+                    cv2.rectangle(img, pt1, pt2, (255, 0, 0), 3, 8, 0)
+
+        cv2.imshow("compressed_result", img)
+        cv2.waitKey(6)		
+		
+		
     rospy.init_node('rosfacedetect')
     rospy.Subscriber(options.topic, sensor_msgs.msg.Image, detect_and_draw)
+	rospy.Subscriber(options.ctopic, sensor_msgs.msg.CompressedImage, compressed_detect_and_draw)
```
vrabaud: indentation

opencv_tests/nodes/source.py
```diff
@@ -71,12 +75,12 @@ def spin(self):
         ball_yv = -ball_yv
 
       self.pub.publish(cvb.cv2_to_imgmsg(cvim))
-
+	  self.pub_compressed.publish(cvb.cv2_to_compressed_imgmsg(cvim))
```
vrabaud: indentation

vrabaud:
Sorry, I was out of town.
We now have a Travis job so please fix it. And please do not use tabs, cf http://wiki.ros.org/PyStyleGuide#Coding_Style
Thx.

talregev:
Hi @vrabaud,
Thanks for the feedback, I fix all the issues you tell me.
But still the test enumerants.py doesn't pass. it fell on line 61.
After compressed and de-compressed the image (from jpg) the shape it not the same.
What should we do?

vrabaud:
Well, that's why we have tests :) If they don't pass, there are errors in your code.

cv_bridge/python/cv_bridge/core.py
```diff
@@ -90,6 +93,46 @@ def encoding_to_cvtype2(self, encoding):
...
+
+        If the image only has one channel, the shape has size 2 (width and height)
+        """
+        import cv2
+        import numpy as np
+
+        str_msg = cmprs_img_msg.data      
+        buf = np.ndarray(shape=(1, len(str_msg)),
+                          dtype=np.uint8, buffer=cmprs_img_msg.data)
+        im = cv2.imdecode(buf, cv2.IMREAD_COLOR)
```
vrabaud: you should not assume it is a color image. Use cv2.IMREAD_ANYCOLOR no ?

cv_bridge/python/cv_bridge/core.py
```diff
@@ -90,6 +93,46 @@ def encoding_to_cvtype2(self, encoding):
...
+        Otherwise desired_encoding must be one of the standard image encodings
+
+        This function returns an OpenCV :cpp:type:`cv::Mat` message on success, or raises :exc:`cv_bridge.CvBridgeError` on failure.
+
+        If the image only has one channel, the shape has size 2 (width and height)
+        """
+        import cv2
+        import numpy as np
+
+        str_msg = cmprs_img_msg.data
```
vrabaud: trailing white spaces, please use a linter (e.g. pylint)

cv_bridge/python/cv_bridge/core.py
```diff
@@ -132,6 +175,45 @@ def imgmsg_to_cv2(self, img_msg, desired_encoding = "passthrough"):
...
+        if not isinstance(cvim, (np.ndarray, np.generic) ):
+            raise TypeError('Your input type is not a numpy array')
+        cmprs_img_msg = sensor_msgs.msg.CompressedImage()
+        cmprs_img_msg.format = dst_format
+        ext_format = '.' + dst_format
+        try:
+            cmprs_img_msg.data = np.array(cv2.imencode(ext_format, cvim)[1]).tostring()
+        except:
+            raise CvBridgeError("format specified as %s is not recognized" % dst_format)
+
```
vrabaud: trailing white spaces.

cv_bridge/python/cv_bridge/core.py
vrabaud: that is probably what's making the test fail.

talregev: I will try.

cv_bridge/python/cv_bridge/core.py
talregev: Do you have online tool for checking? 
This tool is ok?
http://pep8online.com/
It give me many errors.

vrabaud:
Also, be careful with your test, you cannot encode to jpg with 2 channels.

cv_bridge/python/cv_bridge/core.py
```diff
@@ -132,6 +175,45 @@ def imgmsg_to_cv2(self, img_msg, desired_encoding = "passthrough"):
...
+        import numpy as np
+        if not isinstance(cvim, (np.ndarray, np.generic) ):
+            raise TypeError('Your input type is not a numpy array')
+        cmprs_img_msg = sensor_msgs.msg.CompressedImage()
+        cmprs_img_msg.format = dst_format
+        ext_format = '.' + dst_format
+        try:
+            cmprs_img_msg.data = np.array(cv2.imencode(ext_format, cvim)[1]).tostring()
+        except:
+            raise CvBridgeError("format specified as %s is not recognized" % dst_format)
```
vrabaud: you should probably keep the OpenCV error message.

cv_bridge/python/cv_bridge/core.py
vrabaud: it's more for later contributions, what you have right now is fine, just be careful about trailing spaces.

cv_bridge/test/enumerants.py
```diff
@@ -46,13 +48,19 @@ def test_encode_decode_cv2(self):
...
+                        newimg2         = cvb_de.compressed_imgmsg_to_cv2(compress_rosmsg)
                         self.assert_(original.dtype == newimg.dtype)
+                        self.assert_(original.dtype == newimg2.dtype)
                         if channels == 1:
                             # in that case, a gray image has a shape of size 2
                             self.assert_(original.shape[:2] == newimg.shape[:2])
+                            self.assert_(original.shape[:2] == newimg2.shape[:2])
                         else:
                             self.assert_(original.shape == newimg.shape)
+                            self.assert_(original.shape == newimg2.shape)
```
vrabaud: Just have another set of for loops, it will make things easier to track. In that new set, just have the number of channels in ([],1,3,4) (ignore two for jpg).
Also, you are erasing the first set of fmts so that's why you should have a second set (I will fix the tests to use fmts as they don't for now).

talregev: Change all basic linter errors (also for previous tests).

> Just have another set of for loops, it will make things easier to track. In that new set, just have the number of channels in ([],1,3,4) (ignore two for jpg).

Done.

> Also, you are erasing the first set of fmts so that's why you should have a second set (I will fix the tests to use fmts as they don't for now).

Done.

change from IMREAD_COLOR to IMREAD_ANYCOLOR.

Still enumerants.py are falling. 
Maybe if you compressed an image and de-compressed it not identical? 
Because I check the code. it send and receive compressed images. The pics are ok for
the eye, but maybe the shape it not the same?

@vrabaud  Thanks for all feedbacks.

talregev:
@vrabaud now tests are pass.
I am not sure how to test 4 channels.
If there is no more comments, I will squash the commits.

vrabaud:
Not all formats support 4 channels. Jpg and BMP do not. PNG does. Maybe you can skip four channels. 1 and 3 are enough for now.

vrabaud:
Oh, and please squash your commits once you are satisfied with the result. Thx !

talregev:
I squash them! 
@vrabaud

cv_bridge/python/cv_bridge/core.py
```diff
@@ -90,6 +95,46 @@ def encoding_to_cvtype2(self, encoding):
     def encoding_to_dtype_with_channels(self, encoding):
         return self.cvtype2_to_dtype_with_channels(self.encoding_to_cvtype2(encoding))
 
+    def compressed_imgmsg_to_cv2(self, cmprs_img_msg, desired_encoding = "passthrough"):
+        """
+        Convert a sensor_msgs::CompressedImage message to an OpenCV :cpp:type:`cv::Mat`.
+
+        :param img_msg:   A :cpp:type:`sensor_msgs::Image` message
```
vrabaud: please change to cmprs_img_msg and comment properly.

cv_bridge/python/cv_bridge/core.py
```diff
@@ -90,6 +95,46 @@ def encoding_to_cvtype2(self, encoding):
...
+                          dtype=np.uint8, buffer=cmprs_img_msg.data)
+        im = cv2.imdecode(buf, cv2.IMREAD_ANYCOLOR)
+
+        if desired_encoding == "passthrough":
+            return im
+
+        from cv_bridge.boost.cv_bridge_boost import cvtColor2
+
+        try:
+            res = cvtColor2(im, "bgr8", desired_encoding)
```
vrabaud, line 132: Actually, we do not know the original encoding of the image. Let's say it is a JPG with 3 channels. Maybe it is rgb8 or bgr8, or just 8UC3. I'd say you should just remove the desired_encoding and assume it is always passthrough. People can then have their logic outside of the function to convert or not.

vrabaud:
Almost done, I just spotted the logic problem when reading from compressed, let me know what you think.

cv_bridge/python/cv_bridge/core.py
talregev: > please change to cmprs_img_msg and comment properly.

ok

cv_bridge/python/cv_bridge/core.py
talregev, line 132: > Actually, we do not know the original encoding of the image.

what I notice for jpg it always as bgr. 
It will break the api in c++.
https://github.com/ros-perception/vision_opencv/blob/kinetic/cv_bridge/src/cv_bridge.cpp#L435
https://github.com/ros-perception/vision_opencv/blob/kinetic/cv_bridge/src/cv_bridge.cpp#L515

Also read about this in stackoverflow:
http://superuser.com/questions/904978/does-the-jpeg-file-format-support-rgb-data

Also on image_transport_plugins:
https://github.com/ros-perception/image_transport_plugins/blob/indigo-devel/compressed_image_transport/src/compressed_publisher.cpp#L114

talregev:
@vrabaud 

> Almost done, I just spotted the logic problem when reading from compressed, let me know what you think.

I say to keep it that way, because most of the case jpg is in BGR, and to keep the same api in c++.
Also, when you recive jpg picture (or other compressed image) in rviz, it assume it BGR (it has no way to know) and display it as BGR. that why before i send compressed image, i convert it to BGR first.

vrabaud:
Hmm, indeed, that would be against the C++ API.
Then again, it is wrong: what if you send an RGBA PNG ? The code breaks :)

PR #133, opened 2016-05-20
Support compressed Images messages in python for indigo

talregev:
- Add cv2_to_comprssed_imgmsg: Convert from cv2 image to compressed image ros msg.
- Add comprssed_imgmsg_to_cv2:   Convert the compress message to a new image.
- Add compressed image tests.
- Add time to msgs (compressed and regular).

### DIFF
--- a/cv_bridge/python/cv_bridge/core.py
+++ b/cv_bridge/python/cv_bridge/core.py
@@ -1,6 +1,7 @@
 # Software License Agreement (BSD License)
 #
 # Copyright (c) 2011, Willow Garage, Inc.
+# Copyright (c) 2016, Tal Regev.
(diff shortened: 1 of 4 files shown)


talregev:
Hi @vrabaud, I also make this change for kinetic branch.
Change the comment of the commit to be more nice and read.
Can you check and verify the code?
Do you have some comments for improvement? 

Tal.

vrabaud:
Hi, thx for that very useful pull request. I will make all my comments on Kinetic (just because I started there) and will then backport here. So closing for now just not to create confusion.

talregev:
can you open this and close kinetic? because here is automatic tests that i want to check with ...

PR #136, opened 2016-06-07
re-enable travs within docker container

k-okada:


vrabaud:
Is that something standard in ROS for Kinetic on Travis ? would be nice to document it

k-okada:
since travis only support 14.04, if you want to test against 16.40, I think using docker is the only solution.
@130s is pushing to use https://github.com/ros-industrial/industrial_ci as standard (ex. https://github.com/ros-controls/ros_controllers/pull/225 )
But I do not have enough time to follow external repository to keep up with changes, so I believe this PR is best solution for this moment.

vrabaud:
That is a very good one, thx !

PR #141, opened 2016-07-23
Adjust to arbitrary image channels like 32FC40

wkentaro:
I use 32FC40 image for object classification probabilities in 40
classes.

### DIFF
--- a/cv_bridge/python/cv_bridge/core.py
+++ b/cv_bridge/python/cv_bridge/core.py
@@ -94,7 +94,15 @@ class CvBridge(object):
             raise CvBridgeError(e)
 
     def encoding_to_dtype_with_channels(self, encoding):
-        return self.cvtype2_to_dtype_with_channels(self.encoding_to_cvtype2(encoding))
+        try:
+            return self.cvtype2_to_dtype_with_channels(self.encoding_to_cvtype2(encoding))
+        except CvBridgeError as e:
+            try:
+                vals = re.split('(.+)C(.+)', encoding)
+                dtype, n_channels = self.numpy_type_to_cvtype[vals[1]], int(vals[2])
+                return dtype, n_channels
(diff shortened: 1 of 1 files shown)


vrabaud:
is that really useful ? Does the C version handle it ?

wkentaro:
> Does the C version handle it ?

I have no idea how to handle this on C++ with OpenCV.
Currently, I've implemented on Python with Numpy.

> is that really useful ? 

Currently the usecase is only for object segmentation, but I think this is useful.
Below is the example:

(left) **(a) RawImage** (center) **(b) LabelImage** (right) **(c) ProbabilicImageForTargetObject**
image

The pipeline is

```
[Kinect] -> (a) RawImage -> [SegmentationNode] -> (b) LabelImage (32SC1)
                                              |-> ProbabilicImage (32FC40) -> [ChannelExtraction] -> (c) ProbabilicImage (32FC1)
```

vrabaud:
I submitted a proper patch that works with C++ and has tests.

wkentaro:
Thanks! That what I wanted to do but couldn't.

vrabaud:
C++ is the way to go to keep compatibility with both.

wkentaro:
Yeah, I've tried to change the [`getCvType` function](https://github.com/ros-perception/vision_opencv/commit/35df5c389cb599a15a8cc24d4e0d8d9f087aec14#diff-fb2cbb0564c5ce322e4a56960733ecf4R103), but did not reach the use of `CV_MAKETYPE` properly.
I think I got to know the correspondence between OpenCV type and Numpy dtype.

vrabaud:
I wish OpenCV had that by default .... I also added support for "8U" type support (no need for "8UC1")

wkentaro:
I see. Thank you.

PR #142, opened 2016-07-25
remap with nan border if mat value is float or double

YuOhara:
I think border value should be NaN if the elements of Image are expressed in 32FC.

without this patch, in the process of depth2pointcloud in openni2_launch(or something),  the [0, 0, 0] clouds appears like this.
screenshot from 2016-07-23 18 02 02

It is because of this line.
https://github.com/ros-perception/image_pipeline/blob/indigo/depth_image_proc/include/depth_image_proc/depth_traits.h#L57
only nan values are removed.


vrabaud:
makes sense, thx !

k-okada:
@YuOhara can you create PR targeting to the indigo branch as well?

wkentaro:
@YuOhara Thanks!

PR #144, opened 2016-07-29
Fix getConversionCodes to visualize 8UC1 image with cvtColorForDisplay

wkentaro:


wkentaro:
For https://github.com/ros-perception/image_pipeline/issues/216

vrabaud:
I actually disagree. The error message is right.
What if your 8UC1 is a label image ? What you are doing here will not work.
To satisfy your needs, you can instead add options to CvtColorForDisplayOptions like force_indexing and force_continuous. What do you think ?

wkentaro:
> What if your 8UC1 is a label image ? What you are doing here will not work.

I was thinking about it adding `as_label` option to CvtColorForDisplayOptions, but your suggestion sounds better.
Well, what do force_indexing and force_continuous mean? I'm assuming they are flags as below:
- 8UC and 8UC1 as mono (bool flag force_XXX)
- 8UC and 8UC1 as label  (**!** bool flag force_XXX)

wkentaro:
Any comments please?

vrabaud:
well, actually, for the specific title of your PR, you just want to pass a colomap value in CvtColorForDisplayOptions no ? You should also change the logic to have the colormap option work for any input with 1 channel.

wkentaro:
Closing with https://github.com/ros-perception/image_pipeline/issues/216#issuecomment-259820563

PR #150, opened 2016-09-06
[cv_bridge] Implement cpp conversion of N channel image

wkentaro:
This is cpp version of https://github.com/ros-perception/vision_opencv/pull/141,
which is one for python.

### DIFF
--- a/cv_bridge/include/cv_bridge/cv_bridge.h
+++ b/cv_bridge/include/cv_bridge/cv_bridge.h
@@ -300,9 +300,23 @@ CvImageConstPtr cvtColorForDisplay(const CvImageConstPtr& source,
 /**
  * \brief Get the OpenCV type enum corresponding to the encoding.
  *
- * For example, "bgr8" -> CV_8UC3.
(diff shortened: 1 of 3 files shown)


vrabaud:
Let's not duplicate the API: you should upgrade the API here: https://github.com/ros/common_msgs/blob/jade-devel/sensor_msgs/include/sensor_msgs/image_encodings.h#L130
The encodings listed there are just the common ones but you could add a comment saying that any format of the form "8UC_", "64FC_" are valid too and fix bitDepth and numChannels.

wkentaro:
Sounds nice.

vrabaud:
sounds doable to you ? Please don't use boost regex to get heavier dependencies but it should be fairly small to implement

wkentaro:
Yeah, I will do it when I have time.

wkentaro:
Sent https://github.com/ros/common_msgs/pull/91

wkentaro:
The new commit depends on the PR ros/common_msgs#91.

vrabaud:
please ping me again once common_msgs has been released with the added API. Thx.

wkentaro:
common_msgs is released on shadow-fixed.

PR #152, opened 2016-09-20
Fix "stdlib.h: No such file or directory" errors in GCC-6

ghost:
Building with GCC-6 causes errors such as:

`/usr/lib/gcc/x86_64-pc-linux-gnu/6.2.0/include/g++-v6/cstdlib:75:25: fatal error: stdlib.h: No such file or directory
 #include_next <stdlib.h>`

Including `${catkin_INCLUDE_DIRS}` with`-isystem` has the effect of including
`-isystem /usr/include`.

This breaks building with GCC-6.
See https://gcc.gnu.org/bugzilla/show_bug.cgi?id=70129
and https://github.com/ros/geometry/pull/130

### DIFF
--- a/image_geometry/CMakeLists.txt
+++ b/image_geometry/CMakeLists.txt
@@ -12,7 +12,7 @@ catkin_package(CATKIN_DEPENDS sensor_msgs
 
 catkin_python_setup()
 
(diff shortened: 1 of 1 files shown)


vrabaud:
Thx ! Can't wait to try out GCC6 !

PR #153, opened 2016-09-25
cv_bridge: Add missing test_depend on numpy

de-vri-es:
The `cv_bridge` tests depend on `numpy`, but the `package.xml` does not currently declare this. This PR adds the missing `<test_depend>`.


vrabaud:
Good catch, thx !

PR #154, opened 2016-09-30
address gcc6 build error in cv_bridge and tune

bulwahn:
With gcc6, compiling fails with `stdlib.h: No such file or directory`, as including '-isystem /usr/include' breaks with gcc6, cf., https://gcc.gnu.org/bugzilla/show_bug.cgi?id=70129.

This commit addresses this issue for cv_bridge in the same way it was done in the commit ead421b8 [1] for image_geometry. This issue was also addressed in various other ROS packages.
A list of related commits and pull requests is at https://github.com/ros/rosdistro/issues/12783.

[1] https://github.com/ros-perception/vision_opencv/commit/ead421b85eeb750cbf7988657015296ed6789bcf

### DIFF
--- a/cv_bridge/CMakeLists.txt
+++ b/cv_bridge/CMakeLists.txt
@@ -19,8 +19,7 @@ catkin_package(
 
 catkin_python_setup()
 
-include_directories(SYSTEM ${Boost_INCLUDE_DIRS} ${OpenCV_INCLUDE_DIRS})
-include_directories(include ${catkin_INCLUDE_DIRS})
+include_directories(include ${Boost_INCLUDE_DIRS} ${OpenCV_INCLUDE_DIRS} ${catkin_INCLUDE_DIRS})
 
 if(NOT ANDROID)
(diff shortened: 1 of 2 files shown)


vrabaud:
@Hodorgasm , is there more to fix too? I thought https://github.com/ros-perception/vision_opencv/pull/152 was enough.

ghost:
@vrabaud I have been submitting fixes for GCC-6 related bugs based on bug reports on Gentoo Linux bug tracker ([here](https://bugs.gentoo.org/show_bug.cgi?id=582084)).  Gentoo uses this source to compile several related packages.  Since I'm going package by package, I can't guarantee this is the last related bug report but if I have time I will try to compile the whole thing from git in a sandbox this weekend and see.

ghost:
@vrabaud That should be the last GCC-6 bug from this repo.

vrabaud:
cool, thx @Hodorgasm , I'll make a new release soon.

PR #165, opened 2017-04-19
Remove the last remnants of boost from image_geometry.

clalancette:
All of its functionality can be had from std:: in C++11, so
use that instead.  This also requires us to add the -std=c++11
flag.

Signed-off-by: Chris Lalancette <clalancette@osrfoundation.org>

### DIFF
--- a/image_geometry/include/image_geometry/pinhole_camera_model.h
+++ b/image_geometry/include/image_geometry/pinhole_camera_model.h
@@ -22,7 +22,7 @@ public:
 class PinholeCameraModel
 {
 public:
-  
+
   PinholeCameraModel();
 
   PinholeCameraModel(const PinholeCameraModel& other);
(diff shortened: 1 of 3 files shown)


vrabaud:
thx ! I like it clean.

image_geometry/include/image_geometry/pinhole_camera_model.h
```diff
@@ -270,7 +270,7 @@ class PinholeCameraModel
 
   // Use PIMPL here so we can change internals in patch updates if needed
   struct Cache;
-  boost::shared_ptr<Cache> cache_; // Holds cached data for internal use
+  std::shared_ptr<Cache> cache_; // Holds cached data for internal use
```
vrabaud, line 273: that's actually an ABI breakage but I guess switching to C++11 is a higher priority for ROS2.

clalancette: Yeah, C++11 and no boost is more important for ROS2.

You are right that this is an ABI break; will that be a problem for ROS1?

vrabaud: Not too many packages depend on it so that should be fine.

mikaelarguedas: it looks like this change broke downstream packages because the memory header was not included in this header. See #188 for a proposed fix

PR #168, opened 2017-05-11
fix missing initialization of variable

dirk-thomas:
Fixes ros2/vision_opencv#2.


dirk-thomas:
Nvm. That was not yet it. The variable actually gets populated below with `cv::Rodrigues`.

Note to myself: first retest with the patch - afterwards create PR...

dirk-thomas:
@vrabaud Please see what I ended up doing to make it work on Windows debug (VS 2015): ros2/vision_opencv#6

PR #178, opened 2017-07-06
[cv_bridge][indigo] pass when encoding of in and out are same

knorth55:
when `encoding` of `input` and `output` are same, pass `cvtColor`

### DIFF
--- a/cv_bridge/python/cv_bridge/core.py
+++ b/cv_bridge/python/cv_bridge/core.py
@@ -173,7 +173,8 @@ class CvBridge(object):
         if img_msg.is_bigendian == (sys.byteorder == 'little'):
             im = im.byteswap().newbyteorder()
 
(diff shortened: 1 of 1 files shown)


vrabaud:
No, because that would return the same object and you need to give back a copy.
Plus that case is handled properly on the C++ side: https://github.com/ros-perception/vision_opencv/blob/kinetic/cv_bridge/src/cv_bridge.cpp#L311 (and the Python wrappers call the C++ bindings).

knorth55:
ok, thank you.

vrabaud:
Thx for the idea though, I love when people get into details :) Did you see a speed benefit in doing so btw ?

knorth55:
I tested with 100Hz input, actually there is no change :)

vrabaud:
Great ! No reason to look into it then :) Thx !

PR #183, opened 2017-07-27
specify requied components of opencv

ibtaylor:
now the generated pkgconfig file with contain the correct dependencies.

I was using this from another project and it was slowing down my linking stage.

### DIFF
--- a/cv_bridge/CMakeLists.txt
+++ b/cv_bridge/CMakeLists.txt
@@ -13,7 +13,11 @@ if(NOT ANDROID)
 else()
 find_package(Boost REQUIRED)
 endif()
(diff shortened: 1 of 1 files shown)


vrabaud:
Please get the Travis job to work. You are missing a linkage to imgcodecs too: https://travis-ci.org/ros-perception/vision_opencv/builds/257957534?utm_source=github_status&utm_medium=notification

ibtaylor:
My mistake, the components should probably not have the opencv_ prefix.. 

https://github.com/ibtaylor/vision_opencv/commit/5db32151d68626c18f2add92c6a4bdd8e9fd5605

PR #185, opened 2017-09-19
add version_gte for opencv3

k-okada:
@vrabaud If you'll update opencv3 version as discussed in https://discourse.ros.org/t/opencv-3-3/2674/4, I think we'd better to add 'version_gte' tag so that apt-get install ros-kinetic-cv-bridge also pulls openv3.3 from repository, to avoid API breaking issue between opencv2 and opencv3.


vrabaud:
You're totally right, thx.

PR #188, opened 2017-11-07
missing STL includes

mikaelarguedas:
Follow-up of #165, this should fix (at least partially) the regressions we've seen recently on kinetic (e.g. http://build.ros.org/view/Kbin_uX64/job/Kbin_uX64__depthimage_to_laserscan__ubuntu_xenial_amd64__binary/31)

### DIFF
--- a/image_geometry/include/image_geometry/pinhole_camera_model.h
+++ b/image_geometry/include/image_geometry/pinhole_camera_model.h
@@ -5,7 +5,9 @@
 #include <opencv2/core/core.hpp>
 #include <opencv2/imgproc/imgproc.hpp>
 #include <opencv2/calib3d/calib3d.hpp>
+#include <memory>
 #include <stdexcept>
+#include <string>
 
 namespace image_geometry {
 


clalancette:
Oops.  Sorry about that, thanks for looking into this @mikaelarguedas .  The change looks good to me; the Travis failure looks to be unrelated (it's failing a test for endianness), but bears looking into.  If nobody gets to it before then, I can look into it later this week.

mikaelarguedas:
@matlabbe FYI

matlabbe:
Indeed, same error here (c++11 shared pointer not defined): http://build.ros.org/job/Kbin_uX64__rtabmap_ros__ubuntu_xenial_amd64__binary/78/

Should downstream packages have to explicitly add "`add_compile_options(-std=c++11)`"? In a more general way to keep the downstream package's cmake the same for all distributions, is there a version associated with that [pull request](https://github.com/ros-perception/vision_opencv/pull/165) so I can add or not the c++11 option? Or I should assume that if the compiler c++11 is supported, I should build with it? The best would be that c++11 flag be propagated to dependent packages, if it is feasible in catkin.

mikaelarguedas:
>  is there a version associated with that pull request so I can add or not the c++11 option?

The original PR is targeting the kinetic branch. the code on that branch is used for kinetic and lunar.
Looks like (http://repositories.ros.org/status_page/compare_indigo_jade_kinetic_lunar.html?q=image_geometry) you can add the flag for only version 1.12.5 and higher and keep the same CMake code for all distros.

C++11 compatibility has been required only since kinetic so for any code that is copiled for older ROS distributions you should not compile any code using C++11 features. Then I'd expectthat building with the flag when the compiler supports it is ok as long as the projects you compile followed that guideline.

As of exporting the flag that's bit of a tricky one because you don't want to force the flag on downstream packages (most recent versions of gcc will compile with -std=gnu++11 or -std=gnu++14 by default)

mikaelarguedas:
retriggering travis...

mikaelarguedas:
> the Travis failure looks to be unrelated (it's failing a test for endianness), but bears looking into. If nobody gets to it before then, I can look into it later this week.

Given that the failing package and the one modified in this PR don't depend on each other I'm pretty confident that this PR has no relation with the test failure.
I couldnt reproduce the test failure locally so I re-triggered travis in the hope of having a green check on this PR but no luck..

vrabaud:
Thx !

mikaelarguedas:
Thanks @vrabaud! 
Note that gcc<6.0 (the ones in xenial and jessie) still use gnu++98 as the default flag so any downstream package including this header (in kinetic and lunar) needs to pass the c++11 flag explicitly. We could also revert #165 and release it only in melodic where every platform will have gcc > 6.0

mikaelarguedas:
I am personally in favor of the latter as the change doesnt fix any bug and that downstream packages will not have to make any change (and will be able to use the default gnu++14 rather than overwriting the defaut to use c++11).
@clalancette @vrabaud thoughts?

vrabaud:
Fixed in https://github.com/ros-perception/vision_opencv/commit/0b36ea7a6f40fc05d54e3e6871933129687d5cc4
Basically, the message headers are included and therefore the boost/shareds_ptr.hpp or <memory> whether we are on ROS 1 or 2.

PR #211, opened 2018-04-13
remove the inherent boost files

gaoethan:
Those locally inherent files taken from Boost at https://github.com/boostorg/endian.git is unnecessary both in ROS2 and once Boost version higher than 1.58 or above which has the default support


gaoethan:
this is a part of the initial porting, I merge it quickly for the follow-up PRs, please let me if any problem, thanks a lot !

PR #216, opened 2018-05-29
Add apache 2.0 license which aligns to ros2

gaoethan:
Include Apache 2.0 license to align with ros2

Signed-off-by: Ethan Gao <ethan.gao@linux.intel.com>

PR #220, opened 2018-06-01
Revert the license to keep its original

gaoethan:
Revert the license which was aligned to Apache 2.0 in #216  to keep its original


gaoethan:
@mikaelarguedas I've already reverted all the license to its original, please kindly help give a simple review and let me if any problem. thanks !

mikaelarguedas:
This looks good :+1:
Thanks @gaoethan for iterating on this.

Note: In general we add new copyright notice when a significant feature / behavioral change is being made. By the look of it, most of the files changes are about style (for example `rgb_colors.h`) and thus would not qualify to add a new copyright holder.
We can keep it as is for now, but we should keep it in mind in the future.

Piggy-backing on the remark about style changes, please consider (when you have a chance) addressing [the remark](https://github.com/ros2/vision_opencv/issues/13#issuecomment-393461239) about keeping the diff minimal with the ROS1 version, by either submitting style changes upstream or not applying them to the `ros2` branch

mikaelarguedas (approved):
lgtm

PR #223, opened 2018-06-14
backprojection of list of pixels in projectPixelTo3dRay

christian-rauch:
By using `numpy.sqrt()` instead of `math.sqrt()` the backprojection method `projectPixelTo3dRay()` can be used on a list of `uv` coordinates. This makes the call more efficient for backprojecting multiple pixels.

### DIFF
--- a/image_geometry/src/image_geometry/cameramodels.py
+++ b/image_geometry/src/image_geometry/cameramodels.py
@@ -136,7 +136,7 @@ class PinholeCameraModel:
         """
         x = (uv[0] - self.cx()) / self.fx()
         y = (uv[1] - self.cy()) / self.fy()
-        norm = math.sqrt(x*x + y*y + 1)
+        norm = numpy.sqrt(x*x + y*y + 1)
         x /= norm
         y /= norm
         z = 1.0 / norm


christian-rauch:
@mjcarroll Can you have a look at this?

image_geometry/src/image_geometry/cameramodels.py
```diff
@@ -136,7 +136,7 @@ def projectPixelTo3dRay(self, uv):
         """
         x = (uv[0] - self.cx()) / self.fx()
         y = (uv[1] - self.cy()) / self.fy()
-        norm = math.sqrt(x*x + y*y + 1)
+        norm = numpy.sqrt(x*x + y*y + 1)
```
mjcarroll, line 139: I haven't had a chance to verify this myself, but doesn't this change the return type to a `np.array`?  Would that have any downstream impact on users?

christian-rauch: Yes, it changes the return type to a `numpy.array`. It would break code that explicitly assumes the `float` return type instead of a `list` or `nump.array`. If you use the function as before, it will return values with shape `(1,)` instead of a single float.

Do you think this will break code? If so, I could return a POD `float` if the dimension is 1, and a `numpy.array` otherwise.

christian-rauch:
@mjcarroll Any news on this? Can this go ahead or is there something blocking this change?

ijnek:
@christian-rauch I've recently become the maintainer of this repo, and I'm going through all PRs and figuring out which ones are important. Sorry it took this long to get a comment back from a maintainer.

Considering the changes in this PR are breaking changes, I'm hesitant on merging this PR, and hence going to close it out. If you believe these changes are still relevant (probably in ros2), may I kindly ask you to open another PR against ``ros2``, along with some sort of deprecation strategy if possible (because the return type changes). If you think these changes should be merged into ``noetic``, then feel free to comment here and we can re-open it.

PR #231, opened 2018-07-03
fix the build warning from inherent import numpy

gaoethan:
There is build warning to complain the following:

```
In file included from /workspace/codebase/vision_opencv/cv_bridge/src/module_opencv3.cpp:3:0:
/workspace/codebase/vision_opencv/cv_bridge/src/module.hpp: In function 'int do_numpy_import()':
/workspace/codebase/vision_opencv/cv_bridge/src/module.hpp:37:3: warning: converting to non-pointer type 'int' from NULL [-Wconversion-null]
   import_array();
   ^~~~~~~~~~~~
/workspace/codebase/vision_opencv/cv_bridge/src/module.hpp:38:1: warning: control reaches end of non-void function [-Wreturn-type]
 }
 ^
/workspace/codebase/vision_opencv/cv_bridge/src/module_opencv3.cpp: At global scope:
/workspace/codebase/vision_opencv/cv_bridge/src/module_opencv3.cpp:82:19: warning: 'PyObject* failmsgp(const char*, ...)' defined but not used [-Wunused-function]
 static PyObject * failmsgp(const char * fmt, ...)
                   ^~~~~~~~
In file included from /workspace/codebase/vision_opencv/cv_bridge/src/module.cpp:36:0:
/workspace/codebase/vision_opencv/cv_bridge/src/module.hpp: In function 'int do_numpy_import()':
/workspace/codebase/vision_opencv/cv_bridge/src/module.hpp:37:3: warning: converting to non-pointer type 'int' from NULL [-Wconversion-null]
   import_array();
   ^~~~~~~~~~~~
/workspace/codebase/vision_opencv/cv_bridge/src/module.hpp:38:1: warning: control reaches end of non-void function [-Wreturn-type]
 }
 ^
```

But it's derived from the inherent definition for `import_array();` of numpy, so... fix this known warning to avoid error while package installation.

### DIFF
--- a/cv_bridge/src/CMakeLists.txt
+++ b/cv_bridge/src/CMakeLists.txt
@@ -34,6 +34,9 @@ if(PYTHON_VERSION_MAJOR VERSION_EQUAL 3)
   add_definitions(-DPYTHON3)
 endif()
 
+#(TODO gaoethan: address the known warning in more elegant way. this is limited by
+# inherent system definiion to import numpy in module.hpp
+add_compile_options(-Wno-conversion-null -Wno-unused-function -Wno-return-type)
 if(OpenCV_VERSION_MAJOR VERSION_EQUAL 3)
   add_library(${PROJECT_NAME}_boost module.cpp module_opencv3.cpp)
 else()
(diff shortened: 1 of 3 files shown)


mikaelarguedas:
Would it be possible to ignore the warnings only on the `import_array()` lines using pragmas rather than ignoring these warnings for the entire package?
Something similar to: https://github.com/ros/class_loader/blob/f32c278cbac6fa8d16e83bc0ea077399eef8647e/include/class_loader/multi_library_class_loader.hpp#L43-L50

gaoethan:
yes. that's what I intend to narrow down, add_compile_options() is temporary for quick fix, thanks !

PR #233, opened 2018-07-10
use 2 space indentations and comply with uncrustify 0.67

mikaelarguedas:
Connects to ament/uncrustify#24


gaoethan:
@mikaelarguedas lgtm, thanks for your improvement !

PR #235, opened 2018-07-19
Fix buffer type error in ROS2

### LINKED ISSUE #213: buffer type issue of cv_bridge in ROS2 branch
@gaoethan Hi , Thanks for your upstream in ros2 branch , i have succeed in building and import cv_bridge in my own python tests . 
But during the tests , i found that in the function : "imgmsg_to_cv2"  of the file /vision_opencv/cv_bridge/python/cv_bridge/core.py , the code is like this : 

`im = np.ndarray(shape=(img_msg.height, img_msg.width, n_channels),
                           dtype=dtype, buffer=img_msg.data)`

and it will report an error since the img_msg.data is of type "list" and the buffer needs type "Bytes"

i change the img_msg.data to np.array(img_msg.data) and i succeed in using the bridge .

So i am wondering is there a better way to solve this bug ? Or have I missed some points in the code ?

Looking forward to your opinion , thanks !

Besides , i found the opencv_tests in the ros2 branch is still using catkin , but in your [another](https://github.com/gaoethan/vision_opencv/tree/ros2_cv_bridge) gits i found an ament-builded code , so i think you might missed merge the right code in opencv_tests .

pantor:
This fixes #213 for me.

### DIFF
--- a/cv_bridge/python/cv_bridge/core.py
+++ b/cv_bridge/python/cv_bridge/core.py
@@ -169,7 +169,9 @@ class CvBridge(object):
         dtype, n_channels = self.encoding_to_dtype_with_channels(img_msg.encoding)
         dtype = np.dtype(dtype)
         dtype = dtype.newbyteorder('>' if img_msg.is_bigendian else '<')
-        buf = np.asarray(img_msg.data, dtype=dtype)
+
+        buf = np.asarray(img_msg.data, dtype=dtype) if isinstance(img_msg.data, list) else img_msg.data
+
         if n_channels == 1:
             im = np.ndarray(shape=(img_msg.height, img_msg.width),
                             dtype=dtype, buffer=buf)
--- a/cv_bridge/python/cv_bridge/core.py
+++ b/cv_bridge/python/cv_bridge/core.py
@@ -169,12 +169,13 @@ class CvBridge(object):
         dtype, n_channels = self.encoding_to_dtype_with_channels(img_msg.encoding)
         dtype = np.dtype(dtype)
         dtype = dtype.newbyteorder('>' if img_msg.is_bigendian else '<')
+        buf = np.asarray(img_msg.data, dtype=dtype)
         if n_channels == 1:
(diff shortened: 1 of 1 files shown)


gaoethan:
@pantor really thanks for your contribution, generally it has two methods to resolve this issue:

- option 1: make the `img_msg.data` is a byte-like object before giving it to `imgmsg_to_cv2()`
- option 2: make it a byte-like object in the` imgmsg_to_cv2()` just as you contributed in this PR.

Maybe the users are prone to forget to do option 1, so it makes sense to common it with option 2. but it's not necessary to convert again if the `img_msg.data` has already been a byte-like object before, so it's better add a condition checking to judge whether it's a byte-like object and then to convert or not in `imgmsg_to_cv2`, this also assure no overhead to those with the option 1 previously. thanks you !

cv_bridge/python/cv_bridge/core.py
```diff
@@ -169,12 +169,13 @@ def imgmsg_to_cv2(self, img_msg, desired_encoding='passthrough'):
         dtype, n_channels = self.encoding_to_dtype_with_channels(img_msg.encoding)
         dtype = np.dtype(dtype)
         dtype = dtype.newbyteorder('>' if img_msg.is_bigendian else '<')
+        buf = np.asarray(img_msg.data, dtype=dtype)
```
gaoethan: It's better add a condition checking to `img_msg.data` before the real conversion to a byte-like object, thanks !

gaoethan (changes requested):

pantor:
Great, I've added a condition that checks for a list instance. I'm not sure if that covers everything a byte-like check would do.

I find option 2 better than 1. In my case, I use ROS2, convert a `cv::Mat` to a `sensor_msgs::msg::Image` in cpp, send it to another package and convert it back to a `cv::Mat` in python. I think cv_bridge should handle all conversion internally.

gaoethan:
lgtm. 
but a nit-pick: make the variable name `buf` be more accurate.

PR #237, opened 2018-07-31
Add python3_opencv as test dependency

gaoethan:


gaoethan:
This PR expects to work with the rosdep dependency enabling within https://github.com/ros/rosdistro/pull/18686

PR #225, opened 2018-06-15
Update CMakeLists.txt

berkaydeniz:
This change addresses the issue on Mac OS, where the compiler fails to find Boost libraries.

PR #239, opened 2018-08-02
Fix finding Boost Python library after its renaming in Boost 1.67

### LINKED ISSUE #215: Cannot locate Boost-Python
The following is for catkin package: 'cv_bridge'
image

I am having the above problem while i have the following:

image

Totally confused what to do.

meyerj:
Fixes https://github.com/ros-perception/vision_opencv/issues/215.

The boost_python library has been [renamed in Boost 1.67](https://www.boost.org/docs/libs/1_67_0/libs/python/doc/html/rn.html) and the [FindBoost.cmake](https://cmake.org/cmake/help/v3.12/module/FindBoost.html) module requires a Python version suffix.

Currently hard-coding `python27` for Python 2 builds, but I assume that's a reasonable assumption. The alternative would be to extract the major and minor version from `PYTHONLIBS_VERSION_STRING`.

### DIFF
--- a/cv_bridge/CMakeLists.txt
+++ b/cv_bridge/CMakeLists.txt
@@ -6,7 +6,19 @@ find_package(catkin REQUIRED COMPONENTS rosconsole sensor_msgs)
 if(NOT ANDROID)
   find_package(PythonLibs)
   if(PYTHONLIBS_VERSION_STRING VERSION_LESS 3)
-    find_package(Boost REQUIRED python)
+    find_package(Boost QUIET)
+    if(Boost_VERSION LESS 106700)
+      find_package(Boost REQUIRED COMPONENTS python)
+    else()
+      # The boost_python library has been renamed in Boost 1.67 and the FindBoost.cmake
+      # module requires a Python version suffix:
+      #
+      # References:
+      # - https://www.boost.org/docs/libs/1_67_0/libs/python/doc/html/rn.html
+      # - https://cmake.org/cmake/help/v3.12/module/FindBoost.html
+      #
+      find_package(Boost REQUIRED COMPONENTS python27)
+    endif()
   else()
     find_package(Boost REQUIRED python3)
   endif()
--- a/cv_bridge/CMakeLists.txt
+++ b/cv_bridge/CMakeLists.txt
@@ -5,10 +5,17 @@ find_package(catkin REQUIRED COMPONENTS rosconsole sensor_msgs)
 
 if(NOT ANDROID)
   find_package(PythonLibs)
-  if(PYTHONLIBS_VERSION_STRING VERSION_LESS 3)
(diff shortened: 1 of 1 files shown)


meyerj:
Duplicate of #225.

But I think cmake should check for Boost 1.67 and not for Apple. It just happens that macOS+HomeBrew is already using Boost 1.67. But the problem would affect other OS in the same way and on the other side we are also using macOS with pinned older versions of Boost.

NikolausDemmel:
+1, this fixes compilation for me on macos sierra

rossburton:
Obviously this is a bit of a hack for Python 2 (although as there's never going to be a 2.8, it works), but more importantly it totally fails for Python 3 builds.

meyerj:
> Obviously this is a bit of a hack for Python 2 (although as there's never going to be a 2.8, it works), but more importantly it totally fails for Python 3 builds.

Should be fixed in https://github.com/ros-perception/vision_opencv/pull/239/commits/b57fd5b09de2a267d016a22198209669831fc66c, but I did not actually test it yet with Boost 1.67 or greater.

ijnek:
Hi @meyerj, I've recently become a maintainer of this repo, and I'm going through the PRs to figure out which PRs are active or stale. Sorry this PR took a long time.

Considering this PR is against ``kinetic`` which has reached EOL, I'm going to close this out. If you do believe the changes here are necessary in other distros, feel free to retarget another branch. Thanks!

meyerj:
I am not working with macOS anymore those days, but obviously the current version of the [CMake code on branch `noetic`](https://github.com/ros-perception/vision_opencv/blob/noetic/cv_bridge/CMakeLists.txt) is still based on assumptions about the combination of Boost and Python versions that may only hold on specific Linux distrubtions:

https://github.com/ros-perception/vision_opencv/blob/e229d985ac95d5558cca9180343d24ebadfc13f6/cv_bridge/CMakeLists.txt#L6-L18

According to [the documentation](https://cmake.org/cmake/help/v3.16/module/FindBoost.html), whether `find_package(Boost REQUIRED pythonXY)` requires the version suffix `XY` or not actually depends on the Boost version, not on the Python version. Apparently it still somehow works without that for Ubuntu Focal and all supported distributions that also have Python 3.8 or higher (the else branch), but that may be distribution-specific then? Or related to [Boost CMake](https://cmake.org/cmake/help/latest/module/FindBoost.html#boost-cmake) available from version 1.70 on?

Anyway, Python 2 is mostly irrelevant by now, and likely the same holds for Boost versions older than 1.67, so the proposed patch as-is does not add any value anymore. On ROS 2 branches the same issue has already been addresses in different ways (https://github.com/ros-perception/vision_opencv/pull/246, https://github.com/ros-perception/vision_opencv/pull/279, https://github.com/ros-perception/vision_opencv/pull/421), so the best path forward would be to backport those changes back into ROS 1, or to merge them, in order to support a wide range of platforms for both. All that does not depend on the ROS version and applies to ROS 1 and 2 in the same way, I guess.

PR #240, opened 2018-08-14
Add opencv as dependency and change assert_equal with assert

gaoethan:
- Add opencv as dependency
- Replace assert_equal with assert


gaoethan:
This is some optimization to follow-up of https://github.com/ros2/rosdistro/pull/257

PR #241, opened 2018-08-15
Move test dependencies into BUILD_TESTING to avoid build error

gaoethan:
Fix the issue in https://github.com/ros2/rosdistro/pull/260 due to some test dependencies that are find_package'd outside of a `BUILD_TESTING` conditional.

### DIFF
--- a/cv_bridge/CMakeLists.txt
+++ b/cv_bridge/CMakeLists.txt
@@ -46,10 +46,9 @@ add_subdirectory(src)
 # This is overwritten each loop, but which one it points to doesn't really matter.
 set(cv_bridge_lib_dir "$<TARGET_FILE_DIR:${PROJECT_NAME}>")
 
-find_package(ament_lint_auto REQUIRED)
-ament_lint_auto_find_test_dependencies()
-
 if(BUILD_TESTING)
+  find_package(ament_lint_auto REQUIRED)
+  ament_lint_auto_find_test_dependencies()
   add_subdirectory(test)
 endif()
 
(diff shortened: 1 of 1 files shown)


cv_bridge/CMakeLists.txt
```diff
@@ -46,10 +46,9 @@ add_subdirectory(src)
 # This is overwritten each loop, but which one it points to doesn't really matter.
 set(cv_bridge_lib_dir "$<TARGET_FILE_DIR:${PROJECT_NAME}>")
 
-find_package(ament_lint_auto REQUIRED)
-ament_lint_auto_find_test_dependencies()
-
 if(BUILD_TESTING)
+  find_package(ament_lint_auto REQUIRED)
+  ament_lint_auto_find_test_dependencies()
```
mikaelarguedas: This code is already present in the [`test/CMakeLists.txt` file](https://github.com/ros-perception/vision_opencv/blob/ros2/cv_bridge/test/CMakeLists.txt). So having it here as well is redundant, it would be better to remove it.

mikaelarguedas (changes requested):
This code is duplicated so it should be removed from this file rather than moved inside the `BUILD_TESTING` block.

As most linter tests are failing (see log from https://github.com/ros2/rosdistro/pull/260#issuecomment-413094600), we should either fix the test failures or disable the tests will not be fixed (if for example we want to keep the ROS 1 style to keep the `ros2` branch as close to upstream as possible)

gaoethan:
It's removed accordingly, thanks @mikaelarguedas

PR #242, opened 2018-08-16
remove jp2 for test_encode_decode_cv2_compressed

gaoethan:
As JPEG2000 defaults to use JASPER as its codec,
but it's disabled within the opencv library after Ubutnu
Xenial due to security issues, one can enable to test 
after rebuilding opencv library with jasper enabled on
demand while necessary

### DIFF
--- a/cv_bridge/test/conversions.py
+++ b/cv_bridge/test/conversions.py
@@ -1,13 +1,10 @@
-import unittest
-
 from cv_bridge import CvBridge, CvBridgeError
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
discussion about Jasper removal available at https://github.com/ros2/rosdistro/pull/260#issuecomment-413214747 and https://github.com/ros/rosdistro/issues/18173

mikaelarguedas (approved):

PR #243, opened 2018-08-17
Add python3-numpy as build and execution dependency

gaoethan:
`numpy` is necessary not only for test but also a dependency for building and runtime execution.
More details at https://github.com/ros2/rosdistro/pull/260


mikaelarguedas:
`python3-numpy` can now be removed as a `test_depend`.
`test_depend` is only necessary for additional dependencies needed uniquely for testing. As numpy is now listed as a build/exec dependency, the `test_depend` is not necessary anymore.

mikaelarguedas (approved):
:+1:

PR #246, opened 2018-08-21
[ros2] find boost_python on boost 1.58 (<1.65)

mikaelarguedas:
Without this change boost python is not found properly on Ubuntu Xenial as the name of the module has been changed from `python` to `python3` [in Boost 1.65.0](https://github.com/boostorg/python/commit/90829714cc606ecffabb2a151091f10f81e28259) (and Ubuntu Xenial provides Boost 1.58).


gaoethan (approved):
lgtm. thanks @mikaelarguedas to fix this.

PR #256, opened 2018-12-11
enable python support for ros2

ahuizxc:
Modified some files to enable this package support python.


ahuizxc:
@mjcarroll :) please review this PR if you are free thanks.

mjcarroll:
I think it may be better to do it this way: https://github.com/ros-perception/vision_opencv/pull/257  That way we don't have to maintain the `setup.py` file as well.

ahuizxc:
> I think it may be better to do it this way: #257 That way we don't have to maintain the `setup.py` file as well.

Nice job！

mjcarroll:
Superseded by #257

PR #257, opened 2018-12-13
[image_geometry] Enable python package installation

mjcarroll:


dirk-thomas (approved):

image_geometry/CMakeLists.txt
```diff
@@ -22,9 +25,10 @@ add_library(${PROJECT_NAME}
   src/pinhole_camera_model.cpp src/stereo_camera_model.cpp
 )
 target_link_libraries(${PROJECT_NAME} ${OpenCV_LIBRARIES})
-ament_target_dependencies(${PROJECT_NAME}
-  "OpenCV"
-  "sensor_msgs"
+ament_target_dependencies(
+  ${PROJECT_NAME}
+  OpenCV
+  sensor_msgs
```
wjwwood, line 31: Does this actually change anything?

wjwwood (approved):

image_geometry/CMakeLists.txt
dirk-thomas: No.

mjcarroll: I thought the preferred style was without quotes, is that true?

wjwwood: Historically, I don't think so, but maybe the trend has changed 😅

wjwwood: Really doesn't matter though.

chapulina (approved):

PR #274, opened 2019-03-27
Fix Boost Python 2/3 for old and new Boost.

mikepurvis:
This is a simpler version of #239.

I have validated this with Python 2 and Python 3 on Xenial, so I know it works in that scenario (where the proposed solution from #239 does not work for Python 3). And I believe it should work for Boost 1.67+ as well, based on the original fix.

@meyerj @NikolausDemmel @vrabaud

### DIFF
--- a/cv_bridge/CMakeLists.txt
+++ b/cv_bridge/CMakeLists.txt
@@ -3,16 +3,18 @@ project(cv_bridge)
 
 find_package(catkin REQUIRED COMPONENTS rosconsole sensor_msgs)
 
+find_package(Boost QUIET)
 if(NOT ANDROID)
-  find_package(PythonLibs)
-  if(PYTHONLIBS_VERSION_STRING VERSION_LESS 3)
-    find_package(Boost REQUIRED python)
+  if(Boost_VERSION LESS 106700)
+    set(BOOST_PYTHON_COMPONENT_SUFFIX "-py${PYTHON_VERSION_MAJOR}${PYTHON_VERSION_MINOR}")
   else()
-    find_package(Boost REQUIRED python3)
+    set(BOOST_PYTHON_COMPONENT_SUFFIX "${PYTHON_VERSION_MAJOR}${PYTHON_VERSION_MINOR}")
   endif()
+  find_package(Boost REQUIRED COMPONENTS python${BOOST_PYTHON_COMPONENT_SUFFIX})
 else()
-find_package(Boost REQUIRED)
+  find_package(Boost REQUIRED)
 endif()
+
 find_package(OpenCV 3 REQUIRED
   COMPONENTS
     opencv_core
--- a/cv_bridge/CMakeLists.txt
+++ b/cv_bridge/CMakeLists.txt
@@ -3,8 +3,10 @@ project(cv_bridge)
 
 find_package(catkin REQUIRED COMPONENTS rosconsole sensor_msgs)
 
-find_package(Boost QUIET)
 if(NOT ANDROID)
+  find_package(PythonLibs REQUIRED)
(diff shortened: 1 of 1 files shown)


cv_bridge/CMakeLists.txt
```diff
@@ -3,16 +3,18 @@ project(cv_bridge)
 
 find_package(catkin REQUIRED COMPONENTS rosconsole sensor_msgs)
 
+find_package(Boost QUIET)
```
NikolausDemmel: I guess this could be moved one line down, i.e. inside the `if` branch. Probably doesn't matter much.

cv_bridge/CMakeLists.txt
```diff
@@ -3,16 +3,18 @@ project(cv_bridge)
...
 if(NOT ANDROID)
-  find_package(PythonLibs)
```
NikolausDemmel, line 7: The rationale behind removing this is the it gets called implicitly by `find_package(Boost QUIET)`?

cv_bridge/CMakeLists.txt
```diff
@@ -3,16 +3,18 @@ project(cv_bridge)
...
-  if(PYTHONLIBS_VERSION_STRING VERSION_LESS 3)
-    find_package(Boost REQUIRED python)
+  if(Boost_VERSION LESS 106700)
+    set(BOOST_PYTHON_COMPONENT_SUFFIX "-py${PYTHON_VERSION_MAJOR}${PYTHON_VERSION_MINOR}")
```
NikolausDemmel, line 11: The `Boost_VERSION LESS 106700` case looks different from the proposal in #239. I was wondering if that works also on macos with brewed python. But I guess that doesn't matter, since brewed boost is always > 106700 nowadays.

NikolausDemmel (approved):
Thanks Mike. I don't have time to test, but from quick review it looks reasonable. For macos, brewed boost is now > 1.67 (on any version of macos), and that case looks identical to #239 (except for the logic computing the major and minor version digits), so I assume it would work the same.

cv_bridge/CMakeLists.txt
mikepurvis: Oh shoot, it should be. I'm assuming the android environment doesn't have/work with boost. Will fix.

mikepurvis: Oh I guess would have been no harm, but certainly unnecessary to find twice. Fixed.

cv_bridge/CMakeLists.txt
mikepurvis: Oh, no it doesn't. I'm not sure why my testing worked without this, but it definitely should have been there still.

cv_bridge/CMakeLists.txt
mikepurvis: Yeah, I think the main thing for old-Boost is just working with Ubuntu happens to do, which is this.

cv_bridge/CMakeLists.txt
```diff
@@ -4,15 +4,19 @@ project(cv_bridge)
 find_package(catkin REQUIRED COMPONENTS rosconsole sensor_msgs)
 
 if(NOT ANDROID)
-  find_package(PythonLibs)
-  if(PYTHONLIBS_VERSION_STRING VERSION_LESS 3)
-    find_package(Boost REQUIRED python)
+  find_package(PythonLibs REQUIRED)
+  find_package(PythonInterp REQUIRED)
```
mikepurvis, line 8: Need this for `PYTHON_VERSION_*` vars.

mikepurvis:
Xenial and Python 2 are long dead to us; this is no longer relevant.

PR #277, opened 2019-05-22
[Windows][ROS2] Fixed Windows build issues

seanyen:
This change is to address 2 issues found when I am building ROS2 Dashing from source.

* Boost & OpenCV headers are not found at compile time because `${Boost_INCLUDE_DIRS}` & `${OpenCV_INCLUDE_DIRS}` are not in the includes path.

* The imported library of `cv_bridge.lib` is not generated since no functions are dllexport'd from `cv_bridge`, consequently `cv_bridge-utest` fails to build when trying to link against `cv_bridge`. Use `WINDOWS_EXPORT_ALL_SYMBOLS` to make every functions from `cv_bridge` dllexport'd to fix it.

### DIFF
--- a/cv_bridge/src/CMakeLists.txt
+++ b/cv_bridge/src/CMakeLists.txt
@@ -5,6 +5,7 @@ ament_target_dependencies(${PROJECT_NAME}
   "sensor_msgs"
 )
 target_link_libraries(${PROJECT_NAME} ${Boost_LIBRARIES} ${OpenCV_LIBRARIES})
+set_target_properties(${PROJECT_NAME} PROPERTIES WINDOWS_EXPORT_ALL_SYMBOLS ON)
 
 install(TARGETS ${PROJECT_NAME} DESTINATION lib)
 
--- a/cv_bridge/CMakeLists.txt
+++ b/cv_bridge/CMakeLists.txt
@@ -37,7 +37,7 @@ find_package(OpenCV 3 REQUIRED
   CONFIG
 )
 
-include_directories(include)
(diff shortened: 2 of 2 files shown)


seanyen:
It seemed to me that `Dpr__vision_opencv__ubuntu_bionic_amd64` is already broken before this change?

seanyen:
@gaoethan Hope I am pinging the correct owner. Can you help review/merge the change for fixing Windows build?

mjcarroll:
The build is yellow because of the CMake warning w/r/t Boost.  I think that's okay here.

cv_bridge/src/CMakeLists.txt
```diff
@@ -5,6 +5,7 @@ ament_target_dependencies(${PROJECT_NAME}
   "sensor_msgs"
 )
 target_link_libraries(${PROJECT_NAME} ${Boost_LIBRARIES} ${OpenCV_LIBRARIES})
+set_target_properties(${PROJECT_NAME} PROPERTIES WINDOWS_EXPORT_ALL_SYMBOLS ON)
```
mjcarroll, line 8: Should we be doing this or using visibility macros?

mjcarroll:
I think some of the `boost` related problems will be resolved by removing it as a dependency here: https://github.com/ros-perception/vision_opencv/pull/283

I would like to get that landed, then I'll come back and look at this one.

jnoyola:
This would address #289, which I just opened, but I still don't understand why the Boost include only breaks the Windows build.

Also, I was able to build without adding back the OpenCV include, so maybe that is no longer necessary since the PR was opened.

seanyen:
Closed it and in favor of #301

Issue #281, opened 2019-06-25
[ros2] Port boost::python dependency to pybind11


mjcarroll:
The `boost::python` dependency is currently a bit troublesome to maintain in the ROS2 CI configuration across Windows and MacOS.  I'd like to use this as an opportunity to remove the dependency and use `pybind11` instead.

PR #283, opened 2019-07-10
Port boost::python functionality over to pybind11

mjcarroll:
This should potentially alleviate some of our porting headaches to other platforms.

Addresses #281


mjcarroll:
And for giggles:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7518/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3653/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6157/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7346/)

mjcarroll:
With correct library paths for tests:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7524/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3661/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6162/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7353/)

mjcarroll:
macOS fails because of using OpenCV 4.

Windows fails because no `boost/endian` header.

maxtong9:
Hey, I am trying to compile your branch on Windows 10 and I am getting this error:
fatal error C1083: Cannot open include file: 'numpy/ndarrayobject.h': No such file or directory
Am I missing some dependency? I have numpy installed via pip.
Great work on everything you've done! Windows 10 support seems to be lacking still for ROS 2 packages.

mjcarroll:
I have to admit that I haven't sufficiently tested this on Windows.  It may be that a numpy include path isn't getting picked up correctly on Windows?  It would either be in the cmake logic here or in https://github.com/ros2/pybind11_vendor

maxtong9:
Thank you for the quick response! I located where the header should be. I'm going to try and find a way to link it now!

mjcarroll:
Rebuild macOS with OpenCV4 fix: [Build Status](https://ci.ros2.org/job/ci_osx/6499/)

jnoyola:
Just noticed you ran into the same issue as me on Windows  (#289). Would you happen to know why this isn't an issue on the other OS's? Because the Boost include was removed altogether.

jnoyola:
Is removing the version from OpenCV's `find_package` the only change needed to support OpenCV 4? Can we pull that into a separate PR to get it in faster?

klintan:
> Just noticed you ran into the same issue as me on Windows (#289). Would you happen to know why this isn't an issue on the other OS's? Because the Boost include was removed altogether.

I'm on Mac Mojave, and I got that exact issue as well for this branch
`vision_opencv/cv_bridge/src/ndarray_converter.cpp:24:10: fatal error: 'numpy/ndarrayobject.h' file not found`

I think except for that (if that is an issue here somewhere, it would be awesome to get this PR in to vision_opencv, support for opencv4 etc etc.

edit: fixing it for mac is fairly straightforward, just create a symlink for the numpy include headers to the `user/local/include/` folder.

`sudo ln -s /usr/local/Cellar/numpy/1.17.1/lib/python3.7/site-packages/numpy/core/include/numpy /usr/local/include/numpy` 

I'm not sure this is an issue in the CMakeLists file though ? 
Tried to update the cmakelists like here https://github.com/ros2/rosidl_python/issues/66 to no avail.

edit2: 

After building it seems the library is missing, might just be me again, but I had to add 

```
install(TARGETS ${PROJECT_NAME}
        RUNTIME DESTINATION bin
        ARCHIVE DESTINATION lib
        LIBRARY DESTINATION lib
)
```
To the cv_bridge cmakelists.txt file to actually export the library (it seems only the py-bindings get exported)

ijnek:
@mjcarroll Considering this PR has been untouched for almost 3 years, I'm thinking of closing it out, but keeping [the issue](https://github.com/ros-perception/vision_opencv/issues/281) open.

Are you fine with this?

mjcarroll:
Yep.  Still would be nice to do, but it's been on the backburner for long enough that this is no longer relevant.

ijnek:
Thanks @mjcaroll

PR #286, opened 2019-08-13
Fix for case when left.Tx() != 0

danil-tolkachev:
Such a case is possible when using more than two cameras. For example, KITTI projection matrices for color cameras.


clalancette:

Issue #272, opened 2019-03-01
OpenCV4 instead of opencv3


YuTingLiu:
I'm using latest vision of opencv, opencv 4.0.1, is there any plan to update repo to this version?
If you guys don't have a plan, is there any way to modify code to using this version of opencv?
I have already modify src/vision_opencv/cv_bridge/CMakeLists.txt, change:
```
find_package(OpenCV 3 REQUIRED
  COMPONENTS
    opencv_core
    opencv_imgproc
    opencv_imgcodecs
 CONFIG
)
```
to 
```
set(OpenCV_DIR /usr/local/share/opencv4)
find_package(OpenCV 4.0.1 EXACT REQUIRED)`
```
and output shows ROS find OPENCV:
```
system
-- +++ processing catkin package: 'cv_bridge'
-- ==> add_subdirectory(vision_opencv/cv_bridge)
-- Boost version: 1.62.0
-- Found the following Boost libraries:
--   python
-- Found OpenCV: /usr/local (found suitable exact version "4.0.1")
-- Found PythonLibs: /usr/lib/arm-linux-gnueabihf/libpython2.7.so (found suitable version "2.7.13", minimum required is "2.7")
```
But report error when :+1: 
```rospack depends cv_bridge```
```
[rospack] Error: package 'cv_bridge' depends on non-existent package 'opencv3' and rosdep claims that it is not a system dependency. Check the ROS_PACKAGE_PATH or try calling 'rosdep update'
```

sreevan:
I too am facing this, is there any specific workaround or solution for this please? I am on Jetson Tx2 and compiling opencv again would not be possible as i can do that only after flashing to free up space. Please help.

sreevan:
I was able to successfully compile cv_bridge with opencv4 below are the rough notes of what i did:
1) Add set (CMAKE_CXX_STANDARD 11) to your top level cmake
2) In cv_bridge/src CMakeLists.txt line 35 change to if (OpenCV_VERSION_MAJOR VERSION_EQUAL 4)
3) In cv_bridge/src/module_opencv3.cpp change signature of two functions 
  3.1) UMatData* allocate(int dims0, const int* sizes, int type, void* data, size_t* step, int flags, UMatUsageFlags usageFlags) const 
    to
      UMatData* allocate(int dims0, const int* sizes, int type, void* data, size_t* step, AccessFlag flags, UMatUsageFlags usageFlags) const
[...]

PR #288, opened 2019-09-04
add OpenCV4 support

BrutusTT:
- removed the fixed version in the global CMakeList.txt to allow for automatic choice
- added a new case for OpenCV4 + new module_opencv4.cpp with the corresponding modifications

addresses ros-perception/vision_opencv#272


hgaiser:
Isn't this a duplicate of https://github.com/ros-perception/vision_opencv/pull/259 ? Considering opencv2 is not a target anymore of ROS melodic, we can clean up the code (and therefore complexity and make it more future proof) by removing the old module entirely.

clalancette:

PR #292, opened 2019-09-28
fix for endian mismatch issue per boostorg/python PR #218

jwhendy:
Greetings,

I'm on a team maintanining arch linux packages for ROS and someone just ran into [this issue](https://github.com/ros-melodic-arch/ros-melodic-cv-bridge/issues/1). Namely, upon import the user gets:
```
RuntimeError: FATAL: module compiled as little endian, but detected different endianness at runtime
```

In googling around, I found [this](https://github.com/boostorg/python/issues/209) and [this](https://github.com/pni-libraries/python-pni/issues/73) which were handled by PRs [boostorg/python 218](https://github.com/boostorg/python/pull/218) and [python-pni 74](https://github.com/pni-libraries/python-pni/pull/74). Both were the same, and appear based on [this explanation](https://github.com/boostorg/python/pull/218#issue-200647888).

I grepped for `import_array`, added the `return NULL` statement, rebuilt, and I no longer get the import error.

### DIFF
--- a/cv_bridge/src/module.hpp
+++ b/cv_bridge/src/module.hpp
@@ -37,6 +37,7 @@ PyObject* pyopencv_from(const cv::Mat& m);
 static int do_numpy_import( )
 {
     import_array( );
+    return NULL;
 }
 #else
 static void do_numpy_import( )


hgaiser:
I can confirm that this fixes the message of the `RuntimeError`. That error has been there for a while but I always ignored it, but recently with the switch to python3.8 the error got more severe:
```shell
RuntimeError: FATAL: module compiled as little endian, but detected different endianness at runtime
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home/hgaiser/dev/fizyr/ros/cv_bridge/devel_isolated/cv_bridge/lib/python3.8/site-packages/cv_bridge/__init__.py", line 35, in <module>
    exec(__fh.read())
  File "<string>", line 6, in <module>
SystemError: initialization of cv_bridge_boost raised unreported exception
```
So now it was actually crashing code. The proposed change fixes this issue for me.

cv_bridge/src/module.hpp
```diff
@@ -37,6 +37,7 @@ PyObject* pyopencv_from(const cv::Mat& m);
 static int do_numpy_import( )
```
jacobperron, line 37: Should we update the signature to `void *`? This is what I'm proposing in the equivalent `ros2` patch (https://github.com/ros-perception/vision_opencv/pull/346).

ijnek:
Closing, as this has been superceded by https://github.com/ros-perception/vision_opencv/pull/323

PR #295, opened 2019-10-10
Updated fromCameraInfo function to match ROS2 CameraInfo message

luca-della-vedova:
Hi all,

While testing packages that make use of the image_geometry functionality (i.e. the camera_calibration package) I found out that the function fromCameraInfo uses the old definition of ROS message for CameraInfo, where K, D, R and P were uppercase hence it crashed.
I fixed it by changing them to the lowercase version.

While this makes the code "work" I'm open to feedback on whether we should also rename all the class variables to follow the same convention, having more consistency but at the risk of breaking user code that relies on the current one.

Cheers,
/Luca

### DIFF
--- a/image_geometry/image_geometry/cameramodels.py
+++ b/image_geometry/image_geometry/cameramodels.py
@@ -39,15 +39,15 @@ class PinholeCameraModel:
 
         Set the camera parameters from the :class:`sensor_msgs.msg.CameraInfo` message.
         """
-        self.K = mkmat(3, 3, msg.K)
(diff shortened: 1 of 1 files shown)


SteveMacenski (approved):
@wjwwood @mjcarroll can you merge this? The sensor_msgs API changed and this is blocking downstream PRs

wjwwood:
Bump @mjcarroll can you have a look at this?

mjcarroll (approved):

PR #298, opened 2019-10-11
remove check for python3 component of boost

acxz:
Problem:
With boost version 1.71, calling `find_package(Boost REQUIRED python3)` fails. Boost now defaults to `find_package(Boost REQUIRED python)` for both Python versions. This is causing failure to build for the Arch Linux Melodic packages. See [ros-melodic-arch/cv-bridge issue](https://github.com/ros-melodic-arch/ros-melodic-cv-bridge/issues/3).

Solution:
This PR fixes the build process for systems that are using Python 3 and boost >= 1.71. For boost < 1.71 the command for Python 2 remains the same.

Issues:
I expect this PR would break systems that do not have Python 2 and have boost >= 1.71, but since Python 2 is required for ROS melodic via [REP 3](https://www.ros.org/reps/rep-0003.html#melodic-morenia-may-2018-may-2023), this should not be a breaking change for existing platforms.
This may become an issue with the Noetic release, since Noetic targets Python 3 only, but we do not know the Boost version Noetic will target (it may be < 1.71).


meyerj:
See also https://github.com/ros-perception/image_common/pull/134 by the same author.

https://github.com/ros-perception/image_common/pull/143 by @hgaiser might be a more future-proof solution, but probably breaks for older Boost versions that do not understand the versioned Python module name.

acxz:
@mjcarroll or someone else that has maintainer access to this repo, can they take a look at this PR, right now the ArchLinux package is currently patched for this. It would be nice if we have an upstream release with a fix.

acxz:
friendly ping to @vrabaud

acxz:
@mjcarroll can u provide some closure to this PR?

acxz:
@mjcarroll can you take a look at this issue? This is important for getting the codebasee to work on python3 which is a requirement for the noetic release. Maybe I should switch the base branch to noetifc?

acxz:
Closing this issue since https://github.com/ros-perception/vision_opencv/commit/55bf08947a302aec8c83bec1883be1880d56cff5 has been commited,
thx @sloretz!

PR #301, opened 2019-10-15
[Windows][ros2] Export interfaces for Shared Lib on Windows

seanyen:
On Windows, for a shared library, the interfaces must be explicitly decorated by proper `dllimport` and `dllexport` usage. Otherwise, no imported library (.lib) will be generated for the downstream projects to consume at link time.

This pull request is to add the interface export in a portable way provided by `CMake` (instead of directly using MSVC specific `dllimport` and `dllexport`).

This is a rework for https://github.com/ros-perception/vision_opencv/pull/277

### DIFF
--- a/cv_bridge/src/CMakeLists.txt
+++ b/cv_bridge/src/CMakeLists.txt
@@ -1,5 +1,8 @@
 # add library
 add_library(${PROJECT_NAME} cv_bridge.cpp rgb_colors.cpp)
+include(GenerateExportHeader)
+generate_export_header(${PROJECT_NAME} EXPORT_FILE_NAME ${PROJECT_NAME}/${PROJECT_NAME}_export.h)
+target_include_directories(${PROJECT_NAME} PUBLIC "$<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}>")
 ament_target_dependencies(${PROJECT_NAME}
   "OpenCV"
   "sensor_msgs"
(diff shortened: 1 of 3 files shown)


seanyen:
@mjcarroll I created another pull request to only address the `dllexport` and `dllimport` thing (with cleaner solution), and closed the original one.

cv_bridge/src/CMakeLists.txt
```diff
@@ -8,6 +11,10 @@ target_link_libraries(${PROJECT_NAME} ${Boost_LIBRARIES} ${OpenCV_LIBRARIES})
 
 install(TARGETS ${PROJECT_NAME} DESTINATION lib)
 
+install(FILES
+  ${CMAKE_CURRENT_BINARY_DIR}/${PROJECT_NAME}/${PROJECT_NAME}_export.h
+  DESTINATION include/${PROJECT_NAME})
+
```
mjcarroll, line 17: Similar to over here https://github.com/ros2/teleop_twist_joy/pull/14#issuecomment-542380900, removing this section seems to cause it not to install the header, which is a bit unexpected given the documentation

mjcarroll: Ah, I apologize, after rereading I found this line: 
> The resulting file should be installed with other headers in the library.

mjcarroll (approved):
LGTM! Thanks for the contribution.

PR #304, opened 2019-10-17
Updated cv2_to_imgmsg to initialize data in constructor for faster co…

juliantrue:
…nversion in ROS2

Issue: 
ROS2 Image message data assignment causes significant slowdown when interacting with larger images.

Solution proposed: 
Move the data assignment in the to the constructor of the Image message object to bypass expensive assignment operation in ROS2 Image message data type.

Sample Results:
Image size (874, 1164, 3)
Pre-fix: 
img_msg.data = cvim.data # This takes ~1720 ms
Post-fix: 
data = cvim.tostring() # This takes ~1.32ms
img_msg = sensor_msgs.msg.Image(data=data) # This takes ~1.05ms


mjcarroll (approved):

mjcarroll:
This was addressed by #305, thanks for the contribution.

PR #322, opened 2020-03-19
3to4

thebrickhouse21:


ijnek:
Thanks for the PR @thebrickhouse21! I've [recently become the maintainer](https://github.com/ros-perception/vision_opencv/pull/451) of the ros 2 branches, but I've also given access to explicitly merge changes and make releases for the ros 1 branches too, so I'm currently going through the list of PRs.

Sorry that this PR was sitting around for so long. I'm going to close this out, as this would be a breaking change in a distro that has already been released. For others' information, ROS noetic now supports OpenCV4.

PR #321, opened 2020-02-28
OpenCV4 Support. Take 57 

OAkyildiz:
As requested in #308 and #288. The original commit belongs to @solosito, I cherry-picked it into the  [melodic](https://github.com/ros-perception/vision_opencv/tree/melodic) branch. CI test is with Melodic on Bionic Beaver (as opposed to trusty).

>! Original PR's were  a) tested on kinetic w/ Thar, b) pulled from ros2

PR #312, opened 2020-01-28
Python 3 compatibility (Melodic)

sfalexrog:
This P/R does several things:
* adds conditional Python 3 dependencies (as described in #303);
* adds additional Travis job to build and test against Python 3;
* updates tests to work in both Python 2 and 3.

The Travis job is more of a temporary hack, since it cannot use prebuilt ROS packages and has to rebuild everything from source. Hopefully this can be removed once Noetic is out.

PR #308, opened 2019-11-27
Add OpenCV4 support

solosito:
PR of changes from #288 to ros2 branch

PR #323, opened 2020-03-23
Noetic release

ahcorde:
Porting package from Melodic to Noetic

 - Removed Python2 dependencies
 - Updated to Opencv4

For some reason these formats are failing with opencv4: "ppm", "pgm", "pbm", I decided to remove them


mjcarroll:
Since maintainership of the ROS1 side of this has kind of fallen through the cracks, there are several PRs that cover the same ground:

#321, #312, #308, #288, #274, #259

cv_bridge/src/module.hpp
```diff
@@ -33,16 +33,9 @@ int convert_to_CvMat2(const PyObject* o, cv::Mat& m);
 
 PyObject* pyopencv_from(const cv::Mat& m);
 
-#if PYTHON3
 static int do_numpy_import( )
 {
     import_array( );
```
sfalexrog, line 38: Confusingly enough, `import_array` is implemented as a macro - and with newer Python/numpy it actually has a `return NULL` statement inside (I believe [`numpy`'s API generator](https://github.com/numpy/numpy/blob/c1f1bc9ce8e4e2936e80d9bfafc3d8e03237a84b/numpy/core/code_generators/generate_numpy_api.py#L112) may give some insight). What's worse is that the statement is not always called, which leads to some weird side effects - I've had a reasonably recent compiler optimize the whole function call out since not all paths returned a value.

All this is to say, please change the function signature to return `void *` and have a `return nullptr` or something after the `import_array()` :)

ahcorde: https://github.com/ros-perception/vision_opencv/pull/323/commits/060a591116a4d3709031f54f2303e8704e8e4dbb

ahcorde:
@ros-pull-request-builder retest this please

ahcorde:
@ros-pull-request-builder retest this please

ahcorde:
@ros-pull-request-builder retest this please

ahcorde:
@mjcarroll CI is passing here. Who is the mainteiner of this package?

mjcarroll:
I saw you updated some of the travis config, is it expected to pass?

> Who is the mainteiner of this package? 

@vrabaud is, but I think we can get this in to get `noetic` moving along.

mjcarroll (approved):

Stigmaru:
Is OpenCV4 or vision_opencv preinstalled into ROS Noetic or do I need to install OpenCV separately?

ahcorde:
If you install these packages with `apt-get install` then the dependency is installed. If you are compiling them from source use ros-dep. This is the kind of question that you should ask in https://answers.ros.org/questions/

PR #324, opened 2020-04-08
add opencv4 into the dependency list for ROS2

LewisLiuPub:
@gaoethan @ahcorde
I updated code to support OpenCV4 for branch ros2, please have a review.

Signed-off-by: Lewis Liu <wei.zhi.liu@intel.com>

### DIFF
--- a/cv_bridge/CMakeLists.txt
+++ b/cv_bridge/CMakeLists.txt
@@ -37,7 +37,7 @@ else()
 endif()
 
 find_package(sensor_msgs REQUIRED)
-find_package(OpenCV 3 REQUIRED
+find_package(OpenCV 4 REQUIRED
   COMPONENTS
     opencv_core
     opencv_imgproc
--- a/cv_bridge/CMakeLists.txt
+++ b/cv_bridge/CMakeLists.txt
@@ -37,13 +37,23 @@ else()
 endif()
 
 find_package(sensor_msgs REQUIRED)
-find_package(OpenCV 4 REQUIRED
+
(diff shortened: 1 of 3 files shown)


cv_bridge/CMakeLists.txt
```diff
@@ -37,7 +37,7 @@ else()
 endif()
 
 find_package(sensor_msgs REQUIRED)
-find_package(OpenCV 3 REQUIRED
+find_package(OpenCV 4 REQUIRED
```
ahcorde: This line force to OpenCV4 , there is no option for version 3.

You should do something like:

```
find_package(OpenCV 4 QUIET opencv_core opencv_core)
if(NOT OpenCV_FOUND)
  find_package(OpenCV 3 REQUIRED opencv_core opencv_core)
endif()
```

ahcorde (changes requested):
Please review the comment regarding OpenCV

cv_bridge/CMakeLists.txt
LewisLiuPub: It really works, I checked it and uploaded one new patch.
@ahcorde

chapulina (approved):
FWIW this works for me with OpenCV 4 on Focal.

ruffsl:
@mjcarroll @ahcorde did you get a chance to look at this?
I see in https://github.com/ros-perception/vision_opencv/issues/272 you've been discussing about this for foxy release.

ahcorde (approved):

mjcarroll (approved):

PR #330, opened 2020-05-18
[Noetic] Use opencv3 on buster

sloretz:
ROS Noetic uses OpenCV 4 on Focal, but OpenCV 3.2 on Buster: https://www.ros.org/reps/rep-0003.html#noetic-ninjemys-may-2020-may-2025

* Removes debug print added in #323 
* If OpenCV 4 isn't available, try finding 3
* Typedef `AccessFlag` type added in OpenCV 4 if using 3.2
* Use different boost python component names on Buster vs Focal

~~I'm opening this PR a bit early; I'm unable to successfully run tests on my system because the cvtColorForDisplay test fails; but the import is strange. I'm wondering if it will work in the PR job:~~

**Edit** Tests pass; ready for review and hopefully merge and release :rocket:

### DIFF
--- a/cv_bridge/CMakeLists.txt
+++ b/cv_bridge/CMakeLists.txt
@@ -5,7 +5,14 @@ find_package(catkin REQUIRED COMPONENTS rosconsole sensor_msgs)
 
 if(NOT ANDROID)
   find_package(PythonLibs)
-  find_package(Boost REQUIRED python)
+
+  if(PYTHONLIBS_VERSION_STRING VERSION_LESS "3.8")
+    # Debian Buster
+    find_package(Boost REQUIRED python37)
+  else()
+    # Ubuntu Focal
+    find_package(Boost REQUIRED python)
+  endif()
 else()
 find_package(Boost REQUIRED)
 endif()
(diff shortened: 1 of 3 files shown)


sloretz:
As of the latest commit tests pass for be both using Buster and Focal; the cause was a different component name for finding boost::python in Buster versus Focal.

mjcarroll (approved):

cv_bridge/CMakeLists.txt
```diff
@@ -5,11 +5,26 @@ find_package(catkin REQUIRED COMPONENTS rosconsole sensor_msgs)
...
-find_package(OpenCV 4 REQUIRED
+
+set(_opencv_version 4)
+find_package(OpenCV 4 QUIET)
+if(NOT OpenCV_FOUND)
+  message(STATUS "Did not find OpenCV 4, trying OpenCV 3")
+  set(_opencv_version 3)
+endif()
+
+find_package(OpenCV ${_opencv_version4} REQUIRED
```
MatthijsBurgh, line 27: I think this contains a typo and should be `${_opencv_version}` not '${_opencv_version4}'

sloretz: Definitely. Linking to your PR https://github.com/ros-perception/vision_opencv/pull/333

PR #344, opened 2020-07-04
Fix comparison problem, add python suffix for Boost>=1.67

floesche:
The comparisons in the CMakeLists.txt have unexpected results: a `Boost_VERSION` of "1.71" is considered `LESS` than `106500`. This is fixed in 6464b9c.

According to https://cmake.org/cmake/help/latest/module/FindBoost.html?highlight=version%20suffix, Boost Python requires version suffixes. Since `CMakeLists.txt` depends on cmake-3.5, FindPython3 cannot be used, but major and minor versions are instead extracted from `PYTHONLIBS_VERSION_STRING`. This is done in 76feb79.

This allows building the cv_bridge on a freshly installed ubuntu-20.4 with foxy.


allezxandre:
Any chance this could get merged? I see it fails the Bionic build, but without specific CMake code, it appears it's not fixable from the Boost version alone. 

Maybe this could be integrated into a new Foxy branch?

seanyen:
@mjcarroll I confirmed this change is useful and works with Windows build too.

floesche:
That went well… (not, sorry).
I will need to setup a better local test environment before updating this patch.

cv_bridge/CMakeLists.txt
```diff
@@ -18,18 +18,27 @@ if(NOT ANDROID)
     find_package(Boost REQUIRED python)
   else()
     find_package(Boost QUIET)
-    if(Boost_VERSION LESS 106500)
+    if(Boost_VERSION VERSION_LESS 1.65)
```
seanyen, line 21: Should we do `if(Boost_VERSION_STRING VERSION_LESS "1.67")`? For two reasons:
* This check currently doesn't match the CMake `FindBoost` documentation.
* `Boost_VERSION_STRING` will give us more consistent string on the version format.

mjcarroll:
Replaced by #421

PR #342, opened 2020-06-26
Change module.hpp to align with Noetic

mjcarroll:
This makes `module.hpp` match what is in the `noetic` version.

It should also fix #339 by correctly returning a `nullptr` from `do_numpy_import`.

The reason for this is explained in #292.

PR #346, opened 2020-07-09
Add missing return statement in numpy import

jacobperron:
Forward port of #292

This reflects a patch made in https://github.com/boostorg/python/pull/218
I've also updated the return signature to match upstream.

Fixes https://github.com/ros-perception/vision_opencv/issues/339

---

Note, to reproduce the reported error I first built `vision_opencv` with `-DCMAKE_BUILD_TYPE=Release` and then overlayed the [`camera_calibration`](https://github.com/ros-perception/image_pipeline/tree/ros2) package and ran it's tests.

### DIFF
--- a/cv_bridge/src/module.hpp
+++ b/cv_bridge/src/module.hpp
@@ -42,9 +42,10 @@ PyObject * pyopencv_from(const cv::Mat & m);
 #endif
 
 #if PYTHON3
-static int do_numpy_import()
+static void * do_numpy_import()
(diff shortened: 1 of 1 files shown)


mjcarroll:
I also did this in #342 which is looking for a review :crossed_fingers:

jacobperron:
> I also did this in #342 which is looking for a review

Missed that one. Closing in favor of it.

PR #352, opened 2020-08-06
delete __init__ for launch directory

dirk-thomas:
Fixes ros2/ros2#984.

I don't see why this directory should have an `__init__.py` file in the first place?

The `launch` directory is explicitly excluded (https://github.com/ros-perception/vision_opencv/blob/105da5391ea00e261d7ea6ef90aac408877d475f/opencv_tests/setup.py#L9) so the launch file isn't being installed.

But the `nodes` directory is also installed as-is - without using the namespace of the package - which shouldn't be the case since it will collide with the content of a potential package named `nodes`.

@mjcarroll To review and investigate the other inconsistencies.


dirk-thomas:
@mjcarroll Friendly ping.

mjcarroll (approved):

PR #359, opened 2020-09-01
Implement unrectifyImage()

### LINKED ISSUE #355: Unrectify image implementation
Hi I just saw `image_geometry::PinholeCameraModel::unrectifyimage` is not implemented yet.
Are there any plans to implement it?

mintar:
This PR removes the "PinholeCameraModel::unrectifyImage is unimplemented" exception and actually implements the function! :)

Closes #355.

I have also added a unit test. This PR works just fine when no ROI is set in the camera_info message. However, I am unsure how to handle the ROI here. As far as I can tell, since we are doing unrectification, i.e., processing an already rectified image, the ROI is not relevant here.

**Question to the maintainers:** Do we need to handle the ROI for unrectifying images?

* If **yes**, I'll have to address the part marked `// TODO: use K_binned, P_binned, binned_resolution`. I have copied most of the ROI handling code over from `initRectificationMaps`; only the part where these three variables are actually used has to be fixed.
* If **no**, I'll remove all the unnecessary ROI handling code and simplify the PR.

### DIFF
--- a/image_geometry/include/image_geometry/pinhole_camera_model.h
+++ b/image_geometry/include/image_geometry/pinhole_camera_model.h
@@ -107,6 +107,7 @@ public:
    * \return 3d ray passing through (u,v)
    */
   cv::Point3d projectPixelTo3dRay(const cv::Point2d& uv_rect) const;
+  cv::Point3d projectPixelTo3dRay(const cv::Point2d& uv_rect, const cv::Matx34d& P) const;
 
   /**
    * \brief Rectify a raw camera image.
@@ -124,11 +125,13 @@ public:
    * \brief Compute the rectified image coordinates of a pixel in the raw image.
    */
   cv::Point2d rectifyPoint(const cv::Point2d& uv_raw) const;
+  cv::Point2d rectifyPoint(const cv::Point2d& uv_raw, const cv::Matx33d& K, const cv::Matx34d& P) const;
 
   /**
    * \brief Compute the raw image coordinates of a pixel in the rectified image.
    */
   cv::Point2d unrectifyPoint(const cv::Point2d& uv_rect) const;
+  cv::Point2d unrectifyPoint(const cv::Point2d& uv_rect, const cv::Matx33d& K, const cv::Matx34d& P) const;
 
   /**
    * \brief Compute the rectified ROI best fitting a raw ROI.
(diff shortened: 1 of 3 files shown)


mintar:
Closing and reopening to re-trigger build farm tests.

mintar:
@mjcarroll : Any opinions on whether we have to handle the ROI in unrectifyImage? (see my question in the original description)

mjcarroll:
> Question to the maintainers: Do we need to handle the ROI for unrectifying images?

I did a re-read of [REP 104](https://www.ros.org/reps/rep-0104.html) and couldn't justify a strong argument either way.

I'm trying to think of a case where it would be relevant, since as you pointed out, we're already dealing with rectified images.
I suppose you could have a rectified image and contains entirely valid pixels, with an ROI indicating that some of the original raw pixels were discarded.  In that case, if you wanted to project the rectified ROI pixels back in the correct locations in the full resolution, you would have to take the ROI into account.

Does this seem like a valid use case?

mjcarroll:
Alternatively, a datapoint in support of removing would be that we don't deal with the ROI in `unrectifyPoint` currently, either: https://github.com/ros-perception/vision_opencv/blob/noetic/image_geometry/src/pinhole_camera_model.cpp#L377

mintar:
Ok, in that case, I would prefer removing the half-baked ROI support from this PR. That way, `unrectifyImage()` would be consistent with `unrectifyPoint()` (and nobody seems to have missed It there). Would that be okay for you?

mjcarroll:
> Ok, in that case, I would prefer removing the half-baked ROI support from this PR. That way, unrectifyImage() would be consistent with unrectifyPoint() (and nobody seems to have missed It there). Would that be okay for you?

Good with me.

mintar:
So, I've dug into the code again, and found out that `unrectifyPoint` **does** actually deal with ROI and binning correctly, because `K_` and `P_` already include binning and ROI. So I decided to make `unrectifyImage` also handle ROI and binning correctly. It took me longer than I'd like to admit, but I'm sure that now `unrectifyImage` works as expected, and I have the tests to prove it!

@mjcarroll : This PR is now ready for review.

mjcarroll (approved):

mintar:
Thanks for the review!

PR #360, opened 2020-09-02
Add YUV420 conversions to cv_bridge

luca-della-vedova:
This PR introduces support for YUV420 (or I420) conversions to cv_bridge.
YUV420 has 12 bits per pixel, this is represented in cv::Mat objects as a one channel image with 1.5 times the number of rows.

To be tested with / merged after the [matching PR in common_msgs](https://github.com/ros/common_msgs/pull/166).

It was tested with the following pipeline:
`gstreamer pipeline -> OpenCV VideoCap object -> cv_bridge + image_transport -> rqt_image_view`

### DIFF
--- a/cv_bridge/src/cv_bridge.cpp
+++ b/cv_bridge/src/cv_bridge.cpp
@@ -95,6 +95,7 @@ int getCvType(const std::string& encoding)
 
   // Miscellaneous
   if (encoding == enc::YUV422) return CV_8UC2;
+  if (encoding == enc::YUV420) return CV_8UC1;
 
   // Check all the generic content encodings
   boost::cmatch m;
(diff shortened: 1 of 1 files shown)


atasoglou:
Hey @luca-della-vedova,
Thanks for the contribution. We wanted to try this out quickly with noetic (Ubuntu 20.04) but it is not compiling for us. CMake had issues with boost_python3 which we quickly "fixed" but then we noticed also this expects OpenCV 3 while the default for Ubuntu 20.04 is OpenCV 4. Using OpenCV4 it doesn't compile.

Any ideas/clues? Thanks!

**Update:** We just cloned the latest noetic branch and used your changes and it compiles. We now are checking if works.

ijnek:
Hi @luca-della-vedova, I've recently become the maintinaer of this repo, and I'm going through the list of PRs to figure out which ones are still active and important.

I'm going to close this PR out as I believe it is now stale, but if anyone thinks these changes are still relevant, please comment on this PR and we can get it re-opened.

PR #361, opened 2020-09-03
fix for endian mismatch issue per boostorg/python PR #218

duanyutong:


ijnek:
I recently became the maintainer of this repo, and I'm going through the list of PRs and sorting through the active and stale ones.

I'm sorry you had to wait this long for someone to respond to this PR. I'm going to close this PR out as I believe this PR is now stale. Please comment on this PR if that is not the case, and we can re-open it.

duanyutong:
Hi, thanks for attending to this PR. I believe this PR still fixes outstanding bug for melodic and should be merged. We have upgraded to noetic which doesn't suffer from the same issue.

ijnek:
I've reopened this PR. Would you be able to provide some context for the change? The error output, or a link to an existing Github issue would be good. Also, which platform were you on, and using Python2 or Python3?

duanyutong:
We were using python3 with melodic and had the error: `RuntimeError: FATAL: module compiled as little endian, but detected different endianness at runtime`.
This PR was fixing the same issue: https://github.com/boostorg/python/pull/218/files.
And if you look at the main branch of this repo the issue had been fixed: https://github.com/ros-perception/vision_opencv/blob/rolling/cv_bridge/src/module.hpp. It was only melodic that needed the same fix but never got it.

ijnek:
@duanyutong Thanks for the explanation!

ijnek (approved):

Issue #364, opened 2020-09-14
Multiple undefined references in libboost_python3.so


m-morelli:
Hello,
I'm on Debian 10, where I've installed libboost-python1.67-dev. I need to build cv_bridge as a dependency for [navigation2](https://navigation.ros.org/build_instructions/index.html#build-navigation2-dependencies) (foxy version). I have ros2 foxy installed from sources in my machine and properly sourced in the terminal where I do build cv_bridge.

Below [1] is what I get. Any idea of how to solve?

Thanks in advance.
Matteo

[1]
~/nav2_depend_ws$ colcon build --symlink-install 
Starting >>> camera_calibration_parsers
Starting >>> cv_bridge
Finished <<< camera_calibration_parsers [0.37s]                                      
Starting >>> bond
--- stderr: cv_bridge                                                                  
/usr/bin/ld: /usr/lib/x86_64-linux-gnu/libboost_python3.so: undefined reference to `PyUnicode_InternFromString'
/usr/bin/ld: /usr/lib/x86_64-linux-gnu/libboost_python3.so: undefined reference to `PyLong_FromLong'
/usr/bin/ld: /usr/lib/x86_64-linux-gnu/libboost_python3.so: undefined reference to `PyModule_Create2'
/usr/bin/ld: /usr/lib/x86_64-linux-gnu/libboost_python3.so: undefined reference to `PyNumber_InPlaceRemainder'
/usr/bin/ld: /usr/lib/x86_64-linux-gnu/libboost_python3.so: undefined reference to `PySlice_New'
/usr/bin/ld: /usr/lib/x86_64-linux-gnu/libboost_python3.so: undefined reference to `PyList_Sort'
/usr/bin/ld: /usr/lib/x86_64-linux-gnu/libboost_python3.so: undefined reference to `PyObject_SetAttrString'
/usr/bin/ld: /usr/lib/x86_64-linux-gnu/libboost_python3.so: undefined reference to `PyType_IsSubtype'
/usr/bin/ld: /usr/lib/x86_64-linux-gnu/libboost_python3.so: undefined reference to `PyList_Insert'
/usr/bin/ld: /usr/lib/x86_64-linux-gnu/libboost_python3.so: undefined reference to `PyExc_RuntimeError'
/usr/bin/ld: /usr/lib/x86_64-linux-gnu/libboost_python3.so: undefined reference to `PyComplex_Type'
[...]

Issue #318, opened 2020-02-19
Libboost_python error: undefined reference on branch ros2


jdcdp:
I am getting the following error when building vision_opencv on ros2 dashing (debian 10):
```                                                           
--- stderr: cv_bridge                             
/usr/bin/ld: /usr/lib/x86_64-linux-gnu/libboost_python3.so: undefined reference to `PyUnicode_InternFromString'
/usr/bin/ld: /usr/lib/x86_64-linux-gnu/libboost_python3.so: undefined reference to `PyLong_FromLong'
/usr/bin/ld: /usr/lib/x86_64-linux-gnu/libboost_python3.so: undefined reference to `PyModule_Create2'
/usr/bin/ld: /usr/lib/x86_64-linux-gnu/libboost_python3.so: undefined reference to `PyNumber_InPlaceRemainder'
/usr/bin/ld: /usr/lib/x86_64-linux-gnu/libboost_python3.so: undefined reference to `PySlice_New'
/usr/bin/ld: /usr/lib/x86_64-linux-gnu/libboost_python3.so: undefined reference to `PyList_Sort'
/usr/bin/ld: /usr/lib/x86_64-linux-gnu/libboost_python3.so: undefined reference to `PyObject_SetAttrString'
/usr/bin/ld: /usr/lib/x86_64-linux-gnu/libboost_python3.so: undefined reference to `PyType_IsSubtype'
/usr/bin/ld: /usr/lib/x86_64-linux-gnu/libboost_python3.so: undefined reference to `PyList_Insert'
/usr/bin/ld: /usr/lib/x86_64-linux-gnu/libboost_python3.so: undefined reference to `PyExc_RuntimeError'
/usr/bin/ld: /usr/lib/x86_64-linux-gnu/libboost_python3.so: undefined reference to `PyComplex_Type'
/usr/bin/ld: /usr/lib/x86_64-linux-gnu/libboost_python3.so: undefined reference to `PyDict_Copy'
/usr/bin/ld: /usr/lib/x86_64-linux-gnu/libboost_python3.so: undefined reference to `PyEval_CallFunction'
/usr/bin/ld: /usr/lib/x86_64-linux-gnu/libboost_python3.so: undefined reference to `PyDict_Size'
/usr/bin/ld: /usr/lib/x86_64-linux-gnu/libboost_python3.so: undefined reference to `PyUnicode_Type'
[...]

PR #368, opened 2021-01-01
Fix multiple undefined references

homalozoa:
As discussion in [318](https://github.com/ros-perception/vision_opencv/issues/318) and [364](https://github.com/ros-perception/vision_opencv/issues/364). This change should be pulled.


homalozoa:
abandoned

PR #358, opened 2020-08-31
Add equidistant distortion model

mintar:
This PR adds support for the **equidistant** distortion model. In OpenCV, this model is called the [fisheye distortion model](https://docs.opencv.org/3.3.1/db/d58/group__calib3d__fisheye.html). It is based on the following publication:

J. Kannala and S. Brandt (2006). A Generic Camera Model and Calibration Method for Conventional, Wide-Angle, and Fish-Eye Lenses, IEEE Transactions on Pattern Analysis and Machine Intelligence, vol. 28, no. 8, pp. 1335-1340

The equidistant model was listed in [sensor_msgs/distortion_models.h](https://github.com/ros/common_msgs/blob/noetic-devel/sensor_msgs/include/sensor_msgs/distortion_models.h#L47) since Kinetic, but `image_geometry` support was missing until now.

The `equidistant` distortion model is especially useful for high-FOV cameras:

#### Rectified image:

**Left:** `plumb_bob` distortion model
**Right:** `equidistant` distortion model

Notice the distortions in the bottom right and top right corners of the `plumb_bob` image. The camera calibrations for the images above were obtained using Kalibr, and the rectified images were generated using this PR.

--------------------------------

Adding the `equidistant` model was tackled in a number of previous PRs (#184, #299, #306, Intermodalics/vision_opencv#1). To fully implement a new distortion model, the following functions (and *only* those) should be modified: rectifyPoint, rectifyImage, unrectifyPoint, unrectifyImage. The last one is not implemented, so that leaves the first three. Here's an overview of the PRs that I could find that implement the `equidistant` model:

function         | #306 | #184 | #299 | Intermodalics/vision_opencv#1 (2) |
-----------------|------|------|------|------------------------------------|
rectifyPoint     | yes  | no   | no   | no                                 |
rectifyImage **(1)** | yes  | yes  | yes  | yes                                |
unrectifyPoint   | yes  | no   | no   | no                                 |

**(1)** implemented by modifying initRectificationMaps

**(2)** Intermodalics/vision_opencv#1 not only implements the `equidistant` (aka "fisheye_cv") distortion model, but also the `fov` model. That would also be great to have here in a later PR.

So only #306 has implementations of all three functions. Unfortunately it's a bit hard to review because it's a huge diff. Therefore I cleaned up #306 and added set of unit tests. During testing, I noticed several bugs in #306, so I fixed them and decided to submit a separate PR. Thanks to @DavidTorresOcana for his work on #306, upon which this PR is based.

PR #373, opened 2021-02-02
Fix CI

wkentaro:


mjcarroll:
Fixed by #358.  Thanks for the contribution, though.

PR #375, opened 2021-02-04
updated cameramodels.py

paul-shuvo:
For ROS Noetic, at line 10 the following error is thrown:
```
error: File "/opt/ros/noetic/lib/python3/dist-packages/image_geometry/cameramodels.py", line 10, in mkmat
    mat = numpy.matrix(L, dtype='float64')
```
Changed `numpy.matrix` to `numpy.array`. 

In line 119, it throws the following error:
```
error File "/opt/ros/noetic/lib/python3/dist-packages/image_geometry/cameramodels.py", line 120, in project3dToPixel
    dst = self.P * src
ValueError: operands could not be broadcast together with shapes (3,4) (4,1) 
```
Changed it to `dst = self.P @ src` also at line 337  to `dst = self.Q @ src`

### DIFF
--- a/image_geometry/src/image_geometry/cameramodels.py
+++ b/image_geometry/src/image_geometry/cameramodels.py
@@ -7,7 +7,7 @@ import copy
 import numpy
 
 def mkmat(rows, cols, L):
-    mat = numpy.matrix(L, dtype='float64')
+    mat = numpy.array(L, dtype='float64')
     mat.resize((rows,cols))
     return mat
 
@@ -116,7 +116,7 @@ class PinholeCameraModel:
         This is the inverse of :meth:`projectPixelTo3dRay`.
         """
         src = mkmat(4, 1, [point[0], point[1], point[2], 1.0])
-        dst = self.P * src
+        dst = self.P @ src
         x = dst[0,0]
         y = dst[1,0]
         w = dst[2,0]
--- a/image_geometry/src/image_geometry/cameramodels.py
+++ b/image_geometry/src/image_geometry/cameramodels.py
@@ -334,7 +334,7 @@ class StereoCameraModel:
         Note that a disparity of zero implies that the 3D point is at infinity.
         """
         src = mkmat(4, 1, [left_uv[0], left_uv[1], disparity, 1.0])
-        dst = self.Q * src
+        dst = self.Q @ src
         x = dst[0, 0]
         y = dst[1, 0]
         z = dst[2, 0]


mjcarroll:
I'm having trouble understanding what the error is that prompted you to change matrix to array.  Can you provide more details?

paul-shuvo:
Honestly, I don't remember, it was a long time ago. One thing I do remember, for me it worked only when I changed it to `np.array` and the `@` operator. I'll try to check my code again.

ijnek:
Thanks for the PR @paul-shuvo! I've [recently become the maintainer](https://github.com/ros-perception/vision_opencv/pull/451) of the ros 2 branches, but I've also given access to explicitly merge changes and make releases for the ros 1 branches too, so I want to either get this PR merged, or close it out as a stale PR.

Please tell me if you intend to work on these changes still, or otherwise if I don't hear from you I will close the pull request.

paul-shuvo:
hi @ijnek, congrats on becoming the maintainer. As I mentioned earlier, the issue was only fixed when I changed `np.matrix` to `np.array` and `dst = self.P @ src` at line 337 to `dst = self.Q @ src`; the latter one should be the obvious bug whereas the former can be a `numpy` specific issue, which might have been fixed in the latter updates and might not need to be changed if some minimum versioning is enforced, otherwise, I believe it’s better to change it to `np.array` in case some one uses the numpy version that throws the error.

ijnek:
So, I've tried to reproduce the original issue you had with the following script, but it runs fine. So as mjcaroll said, I can't figure out what the original problem you had is. Can you try and reproduce the original problem you had? Without knowing what the issue is, I can't proceed with this PR.

```py
import numpy

def mkmat(rows, cols, L):
    mat = numpy.matrix(L, dtype='float64')
    mat.resize((rows,cols))
    return mat

def project3dToPixel(P, point):
    """
    :param point:     3D point
    :type point:      (x, y, z)
    Returns the rectified pixel coordinates (u, v) of the 3D point,
    using the camera :math:`P` matrix.
    This is the inverse of :meth:`projectPixelTo3dRay`.
    """
    src = mkmat(4, 1, [point[0], point[1], point[2], 1.0])
    dst = P * src
    x = dst[0,0]
    y = dst[1,0]
    w = dst[2,0]
    if w != 0:
        return (x / w, y / w)
    else:
        return (float('nan'), float('nan'))

point = numpy.array([0.0, 0.0, 0.0])
P = mkmat(3, 4, [1.0]*12)
print(project3dToPixel(P, point))
```

paul-shuvo:
Sorry, that was long time ago, I don't have access to that project anymore.

ijnek:
In that case, I'm going to close this PR because there is not enough information to justify the original problem. If anyone identifies the original problem, feel free to open another PR and mention this one against it. Sorry we didnt get to merge it, but thank you for taking the time to open the PR @paul-shuvo , and I hope you are fine with this decision.

PR #379, opened 2021-03-12
OpenCV path changes

premecon2499:
Before Executing Crowd heatmap

execute **sudo pip3 install matplotlib**


mjcarroll:
This doesn't seem like a relevant change?

Issue #349, opened 2020-07-28
Could NOT find Boost (missing: Boost_INCLUDE_DIR python3)


wkoa:
<img width="569" alt="截屏2020-07-28 上午9 25 30" src="https://user-images.githubusercontent.com/21969193/88608409-48a9ce00-d0b4-11ea-8382-2d8b3e905012.png">
In my case, I have failed to execute "**colcon build"**
My environment is :

- MacOS Catalina

- ROS2 Foxy

- Cmake 3.18

- Boost 1.72

Sammysedo:
Did you get any progress on this? I'm stuck at the same issue with Windows 10-built foxy.

wkoa:
> Did you get any progress on this? I'm stuck at the same issue with Windows 10-built foxy.

Not yet

cardboardcode:
+1. Stuck at the same issue as well here.

cardboardcode:
@wkoa @Sammysedo 

Were you guys able to resolve this issue? Been attempting for a week now but to no avail.

Below are details of what I tried so far:

## Environment
1. Windows 10.0.19042 N/A Build 19042
2. Visual Studio Code 2019 Community Edition
3. Boost 1.71.0

## Installation of BOOST
1. Downloaded `boost_1_71_0.zip` from the official `boost.org` website.
2. Extracted to under the administrative C Drive folder `Program Files`.
3. Opened a Visual Studio Native Command Prompt.
4. Navigated to the extracted folder for BOOST 1.71.0 and ran the following command:
```cmd
cd C:\Program Files\boost\boost_1_71_0
bootstrap vc142
.\b2
```

I have looked at #289 which is resolved by an earlier merged pull request and have also verified that the installed BOOST library I have contains a `python` library folder at the following local directory:
> C:\Program Files\boost\boost_1_71_0\libs\python

The `Path` environment variable has also been verified to link to that specific library. Below is a terminal output of the following command:
```cmd
echo %Path% | findstr boost\boost_1_71_0\libs\python
```
#### Terminal Output
```cmd
[...]

PR #390, opened 2021-05-12
fix find boost-python for cv_bridge

flynneva:
this should fix the `cv_bridge` build errors reported for windows #349.

obviously we should try to remove boost going forward for this package.

### DIFF
--- a/cv_bridge/CMakeLists.txt
+++ b/cv_bridge/CMakeLists.txt
@@ -13,23 +13,24 @@ if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
 endif()
 
 if(NOT ANDROID)
-  find_package(PythonLibs)
+  find_package(PythonLibs REQUIRED)
   if(PYTHONLIBS_VERSION_STRING VERSION_LESS 3)
     find_package(Boost REQUIRED python)
   else()
     find_package(Boost QUIET)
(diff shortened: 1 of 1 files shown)


flynneva:
from [this document](https://cmake.org/cmake/help/latest/module/FindBoost.html):
>Note that Boost Python components require a Python version suffix (Boost 1.67 and later), e.g. python36 or python27 for the versions built against Python 3.6 and 2.7, respectively. This also applies to additional components using Python including mpi_python and numpy. Earlier Boost releases may use distribution-specific suffixes such as 2, 3 or 2.7. These may also be used as suffixes, but note that they are not portable.

need to fix the version checks in the cmake file to make it work for Dashing it looks like

flynneva:
for anyone else who ends up here, [boost 1_74_0](https://sourceforge.net/projects/boost/files/boost-binaries/1.74.0/) is the most recent boost release that still uses python 3.8 (compatible with Foxy).

flynneva:
just pushed a fix for the Dashing build that uses an older version of Boost...lets see if it works :)

flynneva:
the dashing build looks like a jenkins docker failure somehow....? not sure whats going on there.

Ace314159:
I'd love if this could get merged. I've experienced this same issue, and I have a similar fix that works for me.

Ace314159:
@mjcarroll Could you please merge this as well?

mjcarroll:
Replaced by #421

PR #394, opened 2021-06-28
Handle padded img msg python

sidrubs:
Images coming off some V4L2 cameras have padding at the end of each line, which results in incorrect matrix generation by the python version of the [imgmsg_to_cv2](https://github.com/ros-perception/vision_opencv/blob/7df66485d409d1fce67c437643e06e95f7800575/cv_bridge/python/cv_bridge/core.py#L147) function.

test_image

The `bytesperline` attribute of the [v4l2_pix_format](https://www.kernel.org/doc/html/v4.14/media/uapi/v4l/pixfmt-v4l2.html) structure specifies the number of bytes between the leftmost pixels in two adjacent lines. This `bytesperline` value is then stored in the `step` attribute of the [Image](https://docs.ros.org/en/api/sensor_msgs/html/msg/Image.html) message.

This fix uses the the `step` value to generate the "correct" shaped matrix, after which the matrix is cropped to the required image `width` and `height`.

### DIFF
--- a/cv_bridge/python/cv_bridge/core.py
+++ b/cv_bridge/python/cv_bridge/core.py
@@ -175,11 +175,11 @@ class CvBridge(object):
         if n_channels == 1:
             im = np.ndarray(shape=(img_msg.height, img_msg.step/dtype.itemsize),
                             dtype=dtype, buffer=img_buf)
-            im = np.ascontiguousarray(im[:img_msg.height, img_msg.width])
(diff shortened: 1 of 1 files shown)


sidrubs:
Am not sure why the `Dpr__vision_opencv__ubuntu_bionic_amd64` test is failing. I don't see how any of the changes in this PR would cause that failure.

PR #395, opened 2021-07-05
remove boost endian

uupks:


cv_bridge/src/cv_bridge.cpp
```diff
@@ -36,10 +36,11 @@
 
 #include <cv_bridge/cv_bridge.h>
 #include <cv_bridge/rgb_colors.h>
-#include <boost/endian/conversion.hpp>
+// #include <boost/endian/conversion.hpp>
```
mjcarroll:
```suggestion
```

mjcarroll:
Thanks for the contribution!

We have made a version of this targeted at the `ros2` branch in #399

PR #423, opened 2022-01-27
Find Python3 if not on Android

sloretz:
@nuclearsandwich FYI

This moves the call to `find_package(Python3` up to the block that happens when NOT on Android.

Should fix https://build.ros2.org/job/Rbin_rhel_el864__cv_bridge__rhel_8_x86_64__binary/60/consoleText


clalancette (approved):
I tested this out on a local Almalinux 8 container.  Before the patch, I was able to reproduce the build failure there.  After the patch, I was not.  Also I double-checked on Ubuntu Focal with this patch in place, and things still build there for me.  Overall, looks good.

mjcarroll (approved):

PR #428, opened 2022-03-22
add fovX and fovY functions in python, cpp

cst0:
Simple functions to provide the field of view for a device given camera intrinsics. Tested on RealSense D435 and seems to work for both stereo and raw image feeds in Python, providing values that are in-line with the documentation. Also ported over to C++.

There's also some minor typo fixes.

### DIFF
--- a/image_geometry/src/image_geometry/cameramodels.py
+++ b/image_geometry/src/image_geometry/cameramodels.py
@@ -113,7 +113,7 @@ class PinholeCameraModel:
 
         Returns the rectified pixel coordinates (u, v) of the 3D point,
         using the camera :math:`P` matrix.
-        This is the inverse of :meth:`projectPixelTo3dRay`.
+        This is the inverse of :math:`projectPixelTo3dRay`.
         """
         src = mkmat(4, 1, [point[0], point[1], point[2], 1.0])
         dst = self.P * src
@@ -132,7 +132,7 @@ class PinholeCameraModel:
 
         Returns the unit vector which passes from the camera center to through rectified pixel (u, v),
         using the camera :math:`P` matrix.
-        This is the inverse of :meth:`project3dToPixel`.
+        This is the inverse of :math:`project3dToPixel`.
         """
         x = (uv[0] - self.cx()) / self.fx()
         y = (uv[1] - self.cy()) / self.fy()
@@ -151,7 +151,7 @@ class PinholeCameraModel:
         :rtype:                 float
 
         Compute delta u, given Z and delta X in Cartesian space.
-        For given Z, this is the inverse of :meth:`getDeltaX`.
+        For given Z, this is the inverse of :math:`getDeltaX`.
         """
         fx = self.P[0, 0]
         if Z == 0:
@@ -168,7 +168,7 @@ class PinholeCameraModel:
         :rtype:                 float
 
         Compute delta v, given Z and delta Y in Cartesian space.
-        For given Z, this is the inverse of :meth:`getDeltaY`.
+        For given Z, this is the inverse of :math:`getDeltaY`.
         """
         fy = self.P[1, 1]
         if Z == 0:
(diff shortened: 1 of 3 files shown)


cst0:
Hi all-- just wondering if there's any interest in this functionality? I've used it a handful of places and felt it made sense to put it in someplace like this rather than continuing to copy it over, but if you feel there's a better repo/project/etc for it please let me know.

(pinging @mjcarroll since you were part of the last PR)

Best,
/Chris

ijnek:
Thanks for the contribution @cst0 ! I've [recently become the maintainer](https://github.com/ros-perception/vision_opencv/pull/451) of the ros 2 branches, but I've also given access to explicitly merge changes and make releases for the ros 1 branches too, so I'm currently going through all the PRs, and figuring out which ones are active and stale.

Considering this PR was only raised this year, can I assume that this PR is still within your interest and should be reviewed?

cst0:
Hey @ijnek! Sure, I'd love to continue with the pull request/review process. Please let me know how much the ROS1 API should reflect the ROS2 API; I can put together the necessary changes on a new branch if there's interest in porting this functionality to newer version.

ijnek:
Sorry, finally got around to this PR!

From my understanding, within ``noetic``, users wouldn't expect ABI breakage. But the changes in this PR would break ABI. (I'm not sure how things are in ROS 1, but that's how it is in ROS 2).

My suggestion (if you've moved to ROS 2), would be to not merge this PR, but look into adding this functionality into ROS 2 rolling. 

I think not breaking ABI is more important than making the new feature available in ``noetic``. What do you think?

cst0:
I agree that there shouldn't be breaking changes, but can you clarify what breaking changes you observe? I've added two functions but haven't removed or modified any existing ones, so I wouldn't expect there to be anything breaking here.

ijnek:
Apologies, you're correct. I thought adding methods would increase the size of the object and hence break ABI, but I was wrong. Adding methods does not increase the size of the object, and so ABI would remain compatible.

image_geometry/src/image_geometry/cameramodels.py
```diff
@@ -278,11 +296,12 @@ def fromCameraInfo(self, left_msg, right_msg):
         # [ 0,  Fy, Cy,  0      ]
         # [ 0,  0,  1,   0      ]
 
-        fx = self.right.P[0, 0]
-        fy = self.right.P[1, 1]
-        cx = self.right.P[0, 2]
-        cy = self.right.P[1, 2]
-        tx = -self.right.P[0, 3] / fx
+        assert self.right.P is not None
+        self._fx = self.right.P[0, 0]
```
ijnek: Why does ``_fx`` and friends have to be a class member?

ijnek (commented):
Hmm, it seems like in C++, you're getting fx from the projection matrix P, but in python, you're getting it from the intrinsic matrix K. Is there a reason for this?

cst0:
It looks like both of these questions are answered by "I should be using fx not K": the C++ implementation already has fx being assigned and available so I can use it, and it looks like in my python porting I was intending to use those class variables (rather than calling `self.right.P[0, 0]`., which would have been more verbose).

Not sure how I mixed that up, but looking at some recorded data I'm seeing P[0,0] == K[0,0], which explains how I missed it in my testing. On Monday I'll be back in the lab and will have access to a wider range of devices. I'll make the change to `fx`, double-check it's reasonable, and report back soon.

image_geometry/src/image_geometry/cameramodels.py
```diff
@@ -1,7 +1,4 @@
-import array
-
 import cv2
-import sensor_msgs.msg
```
ijnek, line 4: Just wondering, could removing these imports accidentally cause unintended breakages for users of this library? Unused dependencies would be good to remove, but probably only in the rolling branch (for ros2), I think...

ijnek:
Rest of the PR looks good now, thanks!

image_geometry/src/image_geometry/cameramodels.py
ijnek: Hmm, let me think about this. Maybe I'm getting this mixed up with #includes in cpp

ijnek: I'm completely wrong on this.

ijnek (approved):

ijnek:
@cst0 would you be able to raise another PR against rolling too?

cst0:
Sure thing! I've done that now.

PR #447, opened 2022-07-31
Link tests against target, and separate tests

ijnek:
This PR separates the tests into different files so they can be run individually (and this is what most ros2 libraries do), and also links against the library and not the individual dependencies.

I'm not too sure if removing ``APPEND_LIBRARY_DIRS`` is going to break the tests on windows or not...


ijnek:
Going to close, and re-open if necessary at a later date.

Issue #262, opened 2019-01-02
change *.h to *.hpp in ros2 branch


xhuan28:
According to the [migration guide](https://index.ros.org/doc/ros2/Migration-Guide/), the suffix of header files should be changed from *.h to *.hpp, but I notice that the the suffixes are still *.h in ros2 branch.

threeal:
Will this issue going to be resolved? as i think renaming the header files will create a problem on other packages that depends on this package.

ijnek:
All .h files have been deprecated in I-turtle, and have removed for J-turtle onwards!

PR #448, opened 2022-07-31
Deprecate .h files in favor of .hpp headers

ijnek:
Resolves: #262

### DIFF
--- a/cv_bridge/include/cv_bridge/cv_bridge.h
+++ b/cv_bridge/include/cv_bridge/cv_bridge.h
@@ -1,0 +1,44 @@
+/*********************************************************************
+* Software License Agreement (BSD License)
+*
+*  Copyright (c) 2011, Willow Garage, Inc,
+*  Copyright (c) 2015, Tal Regev.
+*  Copyright (c) 2018 Intel Corporation.
+*  All rights reserved.
+*
+*  Redistribution and use in source and binary forms, with or without
+*  modification, are permitted provided that the following conditions
+*  are met:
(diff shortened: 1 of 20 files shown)


ijnek:
@wep21 would you be able to review this PR?

ijnek:
@wep21 would you be able to review this PR?

wep21:
I checked all the .h include files are replaced with .hpp include files.

chhtz:
Is there a recommended way to include cv_bridge while staying compatible with older ros distros? (We have a project which ideally should build with Ubuntu20, 22, and 24).
This project made a workaround by checking the distro-name from cmake (this does not look very future-proof, though):
https://github.com/pf-robotics/kachaka-api/pull/149/files

wep21:
@chhtz I guess using `__has_include` is the most easiest way to support multi distro.
https://github.com/pal-robotics/aruco_ros/blob/2eba69e025bb1a7a858792d097270a9f959ab095/aruco_ros/src/simple_double.cpp#L43-L47

chhtz:
@wep21 Indeed I was just starting to write a comment with that solution.

```
#if __has_include("cv_bridge/cv_bridge.hpp")
#  include "cv_bridge/cv_bridge.hpp"
#else
#  include "cv_bridge/cv_bridge.h"
#endif
```

(`__has_include` is defined since gcc 4.9.2 and at least clang 3.0, so it is pretty safe to use at least under Linux. I don't use/did not check MSVC)

PR #454, opened 2022-09-04
Add description of MISSING_Z

ijnek:
MISSING_Z just looks like an unused variable currently, but is actually a value that is made available for users of cv_bridge.

### DIFF
--- a/image_geometry/src/stereo_camera_model.cpp
+++ b/image_geometry/src/stereo_camera_model.cpp
@@ -130,6 +130,9 @@ void StereoCameraModel::projectDisparityTo3d(const cv::Point2d& left_uv_rect, fl
   xyz = XYZ * (1.0/W);
 }
 
+// MISSING_Z is defined as 10000 in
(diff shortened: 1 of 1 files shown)


ijnek:
@wep21 would you be able to review this PR?

wep21:
It looks good to me, because I found the description about missing z in the link.
```
Indicates, whether the function should handle missing values (i.e. points where the disparity was not computed). If handleMissingValues=true, then pixels with the minimal disparity that corresponds to the outliers (see [StereoMatcher::compute](https://docs.opencv.org/3.4/d2/d6e/classcv_1_1StereoMatcher.html#a03f7087df1b2c618462eb98898841345) ) are transformed to 3D points with a very large Z value (currently set to 10000).
```

ijnek:
Thanks. In addition, it is explained in 

https://github.com/ros-perception/vision_opencv/blob/906d326c146bd1c6fbccc4cd1268253890ac6e1c/image_geometry/include/image_geometry/stereo_camera_model.h#L60-L61

ijnek:
@Mergifyio backport foxy galactic humble

PR #480, opened 2022-09-12
Remove opencv2.cpp

ijnek:
Removes opencv2.cpp, and renames opencv3.cpp to opencv4.cpp to more closely match what's been done in ``noetic`` branch. Note that opencv4.cpp does contain an #ifdef for opencv3, so it still works with OpenCV3.

@wep21 Could you review this please?


wep21:
The changes are same as that in https://github.com/ros-perception/vision_opencv/pull/323, so looks good to me.

ijnek:
Thanks @wep21

Issue #485, opened 2022-09-17
``foxy`` and ``galactic`` branch contains breaking changes since last release


ijnek:
When I branched off ``foxy``and ``galactic`` branches, I branched off from the ``ros2`` branch, because that was the dev branch listed in rosdistro for the distributions. However, the ``ros2`` branch contained a significant number of breaking changes, and the ``foxy``and ``galactic`` branches shouldn't have been branched off ``devel``.

Should try and revert all the breaking changes in the two branches to get the ``foxy``and ``galactic`` branches working and make a release.

ijnek:
I've gone through all PRs that were merged since the last release in Foxy and Galactic.

Changes that need to be reverted:

- [ ] #419 
- [ ] #421
- [ ] #422

Changes that don't have to be reverted:

- [x] #352 
- [x] #356 
- [x] #370 
- [x] #392 
- [x] #397 
- [x] #396 
- [x] #399 
- [x] #357 
- [x] #400
- [x] #388 
- [x] #358
- [x] #359
- [x] #378
- [x] #417
- [x] #412
- [x] #326
- [x] #420
- [x] #423
- [x] #418
- [x] #427
- [x] #439
- [x] #252 
- [x] #444 
- [x] #451

Ace314159:
Is there a reason to revert #421 and #422 as well? After the revert, builds do not work when using a Boost 1.67 or newer, as explained in https://github.com/ros-perception/vision_opencv/pull/390#issuecomment-839943369
> from [this document](https://cmake.org/cmake/help/latest/module/FindBoost.html):
> 
> > Note that Boost Python components require a Python version suffix (Boost 1.67 and later), e.g. python36 or python27 for the versions built against Python 3.6 and 2.7, respectively. This also applies to additional components using Python including mpi_python and numpy. Earlier Boost releases may use distribution-specific suffixes such as 2, 3 or 2.7. These may also be used as suffixes, but note that they are not portable.

ijnek:
Hi @Ace314159, sorry for the inconvenience the reversion has caused and thanks for reporting it. Let's get it resolved.

Here's the original reasons:
#419 was reverted because that was the PR that was breaking API in foxy and galactic.
#421 was reverted because it depended a lot on changes from #419
#422 was reverted because it was fixing mistakes in #421.

However, now looking at #390, this is suggesting to me that the foxy/galactic build on windows might fail after the reversions I've made.

[...]

PR #486, opened 2022-09-17
Foxy revert api breakage

ijnek:
This reverts the main PRs that caused API breakage in foxy and galactic, and the associated PRs that depends on the reverted PR. 

Related: #485


ijnek:
@Mergifyio backport galactic

PR #494, opened 2022-10-05
Add new CMake option CV_BRIDGE_ENABLE_PYTHON

light-tech:
To disable building the `cv_bridge` Python support package if desired.

### DIFF
--- a/cv_bridge/CMakeLists.txt
+++ b/cv_bridge/CMakeLists.txt
@@ -12,7 +12,13 @@ if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
   add_compile_options(-Wall -Wextra)
 endif()
 
+option(CV_BRIDGE_DISABLE_PYTHON "Disable building Python bindings" OFF)
+
 if(ANDROID)
+  set(CV_BRIDGE_DISABLE_PYTHON ON)
+endif()
+
+if(CV_BRIDGE_DISABLE_PYTHON)
   find_package(Boost REQUIRED)
   set(boost_python_target "")
 else()
@@ -54,7 +60,7 @@ if(NOT OpenCV_FOUND)
   )
 endif()
 
-if(NOT ANDROID)
+if(NOT CV_BRIDGE_DISABLE_PYTHON)
   ament_python_install_package(${PROJECT_NAME}
     PACKAGE_DIR python/${PROJECT_NAME}
   )
--- a/cv_bridge/src/CMakeLists.txt
+++ b/cv_bridge/src/CMakeLists.txt
@@ -31,7 +31,7 @@ install(FILES
   ${CMAKE_CURRENT_BINARY_DIR}/${PROJECT_NAME}/${PROJECT_NAME}_export.h
   DESTINATION include/${PROJECT_NAME}/${PROJECT_NAME})
 
-if(NOT ANDROID)
+if(NOT CV_BRIDGE_DISABLE_PYTHON)
   Python3_add_library(${PROJECT_NAME}_boost MODULE module.cpp module_opencv4.cpp)
   target_link_libraries(${PROJECT_NAME}_boost PRIVATE
     ${PROJECT_NAME}


ijnek:
Thanks for the PR @light-tech. Could you provide some context for why we need these changes? Does the build fail on your platform without disabling that part of the cmake file? If so, what platform are you using?

light-tech:
Hi @ijnek, I am building an "unpack and run" archive for the Raspberry Pi (c.f. https://github.com/light-tech/TurtleBot3-On-RPi) and I can't seem to get Boost python to work on the platform in any way, probably for the same reason Python was disabled on Android?

ijnek:
Sorry the delay in reply @light-tech .
I'd like to know what the error messages are and if they can't be solved, then we could consider merging this PR.

light-tech:
@ijnek No problem. The issue is the unavailability of Boost Python on many platforms such as

 * macOS: I reported https://github.com/boostorg/python/issues/399 (though I could hacked its build tool to make it work);
 * (_this is probably just me_) Debian 11: probably due to existence of both Python 3.9 and Python 3.10 (unfortunately I don't have access to the machine anymore);
 * iOS: obvious as like Android, it does not have Python (I am building part of ROS2 for iOS at https://github.com/light-tech/ROS2-On-iOS).

Instead of this small hack, I could think of a better option which is to split `cv_bridge` into `cv_bridge_cpp` and `cv_bridge_py` like many other ROS2 projects. But that might sound too much?

ijnek:
Sorry about the delay and thanks for the detailed examples. Down the road, I'd like to remove boost, like https://github.com/ros-perception/vision_opencv/pull/283 tried to. Although not amazing, disabling the python if boost is not available seems like the most sane option here.

May I suggest we replace the ``ANDROID`` flag with ``CV_BRIDGE_DISABLE_PYTHON`` everywhere, and set ``CV_BRIDGE_DISABLE_PYTHON`` to ON if the platform is ANDROID?

light-tech:
@ijnek I made the changes according to your suggestion which is better as the change is minimal. Switching to Pybind11 does appear to be a better choice. I am not familiar with it though.

cv_bridge/CMakeLists.txt
```diff
@@ -12,7 +12,13 @@ if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
   add_compile_options(-Wall -Wextra)
 endif()
 
+option(CV_BRIDGE_DISABLE_PYTHON "Build Python package" OFF)
```
ijnek:
```suggestion
option(CV_BRIDGE_DISABLE_PYTHON "Disable building Python bindings" OFF)
```

cv_bridge/CMakeLists.txt
```diff
@@ -12,7 +12,13 @@ if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
...
+
 if(ANDROID)
+  set(CV_BRIDGE_DISABLE_PYTHON ON)
+endif()
+
+if(CV_BRIDGE_DISABLE_PYTHON)
   find_package(Boost REQUIRED)
```
ijnek, line 22: I see that ``find_package(Boost REQUIRED)`` still runs even if you disable cv bridge here. Is this problematic for your platform or is it fine?

cv_bridge/CMakeLists.txt
light-tech: Thanks for pointing out this. Corrected!

cv_bridge/CMakeLists.txt
light-tech: Boost itself is fine as only `boost_python` and `boost_coroutine` which requires some assembly code are missing.

ijnek: Got it, thanks for the clarification!

ijnek (approved):

ijnek:
Thanks for the contribution @light-tech ! As only bug fixes are ported back (and I think this is more of a feature addition than a bug fix), these changes would be available from ROS 2 Rolling and I-turtle onwards.

PR #495, opened 2022-10-12
Address RHEL buildfailure with Boost 1.66.0

ijnek:
Fixes a problem introduced in https://github.com/ros-perception/vision_opencv/pull/489/files, when boost version is 1.66 (ie. on RHEL).

``python3`` should be the correct component used from Boost, not ``python``.

### DIFF
--- a/cv_bridge/CMakeLists.txt
+++ b/cv_bridge/CMakeLists.txt
@@ -25,7 +25,7 @@ if(NOT ANDROID)
       # Resolution for newer versions:
       #  https://gitlab.kitware.com/cmake/cmake/issues/16391
       set(_Boost_PYTHON3_HEADERS "boost/python.hpp")
-      find_package(Boost REQUIRED python)
+      find_package(Boost REQUIRED python3)
     else()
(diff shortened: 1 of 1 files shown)


ijnek:
@Mergifyio backport foxy

Ace314159 (approved):
Looks good assuming you've verified this fixes the build! I noticed in the other `find_package` calls they had the `COMPONENTS` keyword. Is that not required here as well?

ijnek:
The COMPONENTS keyword could be used, but shouldn't be necessary in this case. Thanks for the review, going to check what the CI failure is now.

ijnek:
CI failure was due to 

Boost_VERSION returns "1.71.0" with CMake version newer than 3.15, whereas in CMake versions before 3.15, it returned "107100". See [CMP0093](https://cmake.org/cmake/help/latest/policy/CMP0093.html#policy:CMP0093) for more. So, with the current condition:

```sh
if(Boost_VERSION LESS 106700)
```

Evaluates True for all values of Boost_VERSION. This should be changed to 

```sh
if(Boost_VERSION_STRING VERSION_LESS "1.67")
```

which should evaluate it correctly.

PR #498, opened 2022-10-30
Ensure dynamic scaling works when given matrix with inf and -inf values.

### LINKED ISSUE #450: Dynamic scaling doesn't work with -inf, nan and inf values
Depth images can contain -inf, nan and inf values as described in REP-117, but dynamic scaling for depth images doesn't work because it considers these values when calculating the min and max of the depth.

-Inf, NaN and Inf should be excluded from the calculation of the maximum and minimum value of the depth image.

ijnek:
[REP-117](https://www.ros.org/reps/rep-0117.html) mentions that inf and -inf are valid values in depth images. However, dynamic_scaling doesn't currently work when such values are in the image as the cv::minMaxLoc method doesn't expect inf and -inf values (it's fine with nan values).

By using a mask, we can ignore inf and non inf values when calling ``cv::minMaxLoc``.

Resolves: #450

Signed-off-by: Kenji Brameld <kenjibrameld@gmail.com>

### DIFF
--- a/cv_bridge/src/cv_bridge.cpp
+++ b/cv_bridge/src/cv_bridge.cpp
@@ -649,7 +649,9 @@ CvImageConstPtr cvtColorForDisplay(
 
   // Perform scaling if asked for
   if (options.do_dynamic_scaling) {
-    cv::minMaxLoc(source->image, &min_image_value, &max_image_value);
+    float inf = std::numeric_limits<float>::infinity();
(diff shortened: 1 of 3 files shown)


ijnek:
@wep21 Could I ask you to review this one please?

wep21:
LGTM

ijnek:
Thanks @wep21 !

PR #505, opened 2023-02-01
Allow users to override encoding string in ROSCvMatContainer

### LINKED ISSUE #504: ROSCvMatContainer assumes encoding string for cv::Mat type CV_8UC3
The implementation of [convert_to_ros_message](https://github.com/ros-perception/vision_opencv/blob/c620dd9210bae632ebb442a63f34e9e0d469c57b/cv_bridge/include/cv_bridge/cv_mat_sensor_msgs_image_type_adapter.hpp#L241-L243) sets the encoding string to `bgr8` if the `cv::Mat` is of type `CV_8UC3`. But in reality the same type, ie 8bit 3 channel, could also be used for `rgb8` encodings.

This is acknowledged in [encoding2mat_type(..)](https://github.com/ros-perception/vision_opencv/blob/c620dd9210bae632ebb442a63f34e9e0d469c57b/cv_bridge/src/cv_mat_sensor_msgs_image_type_adapter.cpp#L38-L49) where both `bgr8` and `rgb8` return type `CV_8UC3`.

When implementing camera drivers using this type adapter, the choice of encoding should be left to the user. The auto-inference of the encoding based on cv::Mat type is useful but having an override in the `ROSCvMatContainer` API would be ideal.

Yadunund:
Fixes #504 

The implementation here relies on using an `std::optional<std::string> encoding_override` argument which can be supplied during instantiation of `ROSCVMatContainer`. If provided, the value will be used, else default to the existing behavior of inferring the encoding from `cv::Mat::type()`.
In order to use `std::optional`, I also had to bump the C++ version to C++17.

Also it looks like `ROSCvMatContainer::get_sensor_msgs_msg_image_copy()` has duplicate code from `convert_to_ros_message()`. I've left a TODO to explore reusing the latter within this function. Happy to implement it if the change is acceptable.

Signed-off-by: Yadunund <yadunund@openrobotics.org>

### DIFF
--- a/cv_bridge/CMakeLists.txt
+++ b/cv_bridge/CMakeLists.txt
@@ -3,9 +3,9 @@ project(cv_bridge)
 
 find_package(ament_cmake_ros REQUIRED)
 
-# Default to C++14
+# Default to C++17
 if(NOT CMAKE_CXX_STANDARD)
-  set(CMAKE_CXX_STANDARD 14)
+  set(CMAKE_CXX_STANDARD 17)
 endif()
 
 if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
--- a/cv_bridge/include/cv_bridge/cv_mat_sensor_msgs_image_type_adapter.hpp
+++ b/cv_bridge/include/cv_bridge/cv_mat_sensor_msgs_image_type_adapter.hpp
@@ -26,6 +26,8 @@
 
 #include "cv_bridge/visibility_control.h"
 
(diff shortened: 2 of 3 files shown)


ijnek:
@Yadunund thanks for raising an issue and the PR too. Could you create an issue as an alternative to a todo comment. The changes itself look good.

Yadunund:
Thanks for the quick review. I've removed the todo comment and opened this ticket instead https://github.com/ros-perception/vision_opencv/issues/506

ijnek:
Thanks for the contribution!

PR #507, opened 2023-02-12
Changed to make provision for using OpenCV 4

Shobuj-Paul:
The compiler will check for OpenCV 4 in the system before defaulting to OpenCV 3.
Have also made a few changes in module_opencv2.cpp to cv::Mat.refcount object, since it was removed in OpenCV 3. Passed a separate integer pointer to manage refcount.

Solves #508


clalancette:

PR #514, opened 2023-07-13
Fixed conversion for 32FC1

ahcorde:
I was trying to visualize a 32FC1 with `image_view` but the conversion was failing, this fix should at least show the image instead of a black image.


ijnek (approved):
Thanks!

PR #515, opened 2023-07-19
rolling use c++17

mosfet80:
rolling use c++17 ->switch to c++17.
removed unused travis file


ijnek (approved):
Thank you for your contribution!

PR #518, opened 2023-11-15
fmt: add lint support for cv_bridge and image_geometry && add auto cmake support

homalozoa:
I did several changes for this repo:

for cv_bridge:
1. Using auto cmake to make cmake files more neater and simplier. So some useless cmakelists are removed from the directory.
2. Turn on all lint support without copyright checking. Because there are some files I don't know how to correct under the ROS 2 copyright rules.
3. Reformat many files to pass all test checkings.

for image_geometry:
1. Using auto cmake to make cmake files more neater and simplier. So some useless cmakelists are removed from the directory.
2. Turn on all lint support.
3. Reformat many files to pass all test checkings.


homalozoa:
emmmm.  There is a little trouble when rosdep finds OpenCV/libopencv-dev. I'll check later.

PR #521, opened 2024-03-01
Support for passing parameters to cv::imencode through "toCompressedImageMsg" function.

emrekuru97:
Support for passing parameters to cv::imencode through `toCompressedImageMsg` function, i need this functionality so that the compression quality can be passed through `toCompressedImageMsg` function, other pepople can pass other options too. Tried to keep current functionality and generated possible other signatures  for the function `toCompressedImageMsg`. Could not find a formatter in the repo so did not formatting, the default ros formatter changed a lot so i let it.

### DIFF
--- a/cv_bridge/include/cv_bridge/cv_bridge.h
+++ b/cv_bridge/include/cv_bridge/cv_bridge.h
@@ -117,10 +117,24 @@ public:
    * can be: jpg, jp2, bmp, png, tif at the moment
    * support this format from opencv:
    * http://docs.opencv.org/modules/highgui/doc/reading_and_writing_images_and_video.html#Mat imread(const string& filename, int flags)
+   * params are the options for cv::imencode.
+   * Default value is empty list.
    */
   sensor_msgs::msg::CompressedImage::SharedPtr toCompressedImageMsg(
-    const Format dst_format =
-    JPG) const;
+    const Format dst_format = JPG,
+    const std::vector<int> &params=std::vector<int>()) const;
+
+  /**
+   * dst_format is compress the image to desire format.
+   * Default value is empty string that will convert to jpg format.
+   * can be: jpg, jp2, bmp, png, tif at the moment
+   * support this format from opencv:
+   * http://docs.opencv.org/modules/highgui/doc/reading_and_writing_images_and_video.html#Mat imread(const string& filename, int flags)
+   * params are the options for cv::imencode.
+   * Default value is empty list.
+   */
+  sensor_msgs::msg::CompressedImage::SharedPtr toCompressedImageMsg(
+    const std::vector<int> &params=std::vector<int>(), const Format dst_format = JPG) const;
 
   /**
    * \brief Copy the message data to a ROS sensor_msgs::msg::Image message.
@@ -136,11 +150,25 @@ public:
    * can be: jpg, jp2, bmp, png, tif at the moment
    * support this format from opencv:
    * http://docs.opencv.org/modules/highgui/doc/reading_and_writing_images_and_video.html#Mat imread(const string& filename, int flags)
+   * params are the options for cv::imencode.
+   * Default value is empty list.
    */
   void toCompressedImageMsg(
     sensor_msgs::msg::CompressedImage & ros_image,
-    const Format dst_format = JPG) const;
+    const Format dst_format = JPG, const std::vector<int> &params=std::vector<int>()) const;
 
+  /**
+   * dst_format is compress the image to desire format.
+   * Default value is empty string that will convert to jpg format.
+   * can be: jpg, jp2, bmp, png, tif at the moment
+   * support this format from opencv:
+   * http://docs.opencv.org/modules/highgui/doc/reading_and_writing_images_and_video.html#Mat imread(const string& filename, int flags)
+   * params are the options for cv::imencode.
+   * Default value is empty list.
+   */
+  void toCompressedImageMsg(
+    sensor_msgs::msg::CompressedImage & ros_image,
+    const std::vector<int> &params=std::vector<int>(), const Format dst_format = JPG) const;
 
   typedef std::shared_ptr<CvImage> Ptr;
   typedef std::shared_ptr<CvImage const> ConstPtr;
--- a/cv_bridge/src/cv_bridge.cpp
+++ b/cv_bridge/src/cv_bridge.cpp
@@ -464,13 +464,19 @@ CvImagePtr cvtColor(
 
 /////////////////////////////////////// CompressedImage ///////////////////////////////////////////
 
-sensor_msgs::msg::CompressedImage::SharedPtr CvImage::toCompressedImageMsg(const Format dst_format)
+sensor_msgs::msg::CompressedImage::SharedPtr CvImage::toCompressedImageMsg(const Format dst_format,const std::vector<int> &params)
 const
 {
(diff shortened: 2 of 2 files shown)


cv_bridge/include/cv_bridge/cv_bridge.h
```diff
@@ -117,10 +117,24 @@ class CV_BRIDGE_EXPORT CvImage
...
    * support this format from opencv:
    * http://docs.opencv.org/modules/highgui/doc/reading_and_writing_images_and_video.html#Mat imread(const string& filename, int flags)
+   * params are the options for cv::imencode.
+   * Default value is empty list.
    */
   sensor_msgs::msg::CompressedImage::SharedPtr toCompressedImageMsg(
-    const Format dst_format =
-    JPG) const;
+    const Format dst_format = JPG,
+    const std::vector<int> &params=std::vector<int>()) const;
```
christianrauch, line 125: There need to be spaces around `=` to follow the style in the rest of the code.  I also suggest using the curly braces initialiser (`{}`) in place of the explicit constructor for easier readability.

Adding a new parameter to a function breaks the ABI, meaning that the new library can not be replaced without recompiling all packages that link it. For backward compatibility, you have to keep a function with the old signature that calls your new function with an empty `params`.

cv_bridge/src/cv_bridge.cpp
```diff
@@ -464,15 +464,21 @@ CvImagePtr cvtColor(
 
 /////////////////////////////////////// CompressedImage ///////////////////////////////////////////
 
-sensor_msgs::msg::CompressedImage::SharedPtr CvImage::toCompressedImageMsg(const Format dst_format)
+sensor_msgs::msg::CompressedImage::SharedPtr CvImage::toCompressedImageMsg(const Format dst_format,const std::vector<int> &params)
```
christianrauch, line 467: Add a space between the end of a parameter and the start of a new one. See the style in other parts of the code.

cv_bridge/include/cv_bridge/cv_bridge.h
```diff
@@ -117,10 +117,24 @@ class CV_BRIDGE_EXPORT CvImage
...
+   * dst_format is compress the image to desire format.
+   * Default value is empty string that will convert to jpg format.
+   * can be: jpg, jp2, bmp, png, tif at the moment
+   * support this format from opencv:
+   * http://docs.opencv.org/modules/highgui/doc/reading_and_writing_images_and_video.html#Mat imread(const string& filename, int flags)
+   * params are the options for cv::imencode.
+   * Default value is empty list.
+   */
+  sensor_msgs::msg::CompressedImage::SharedPtr toCompressedImageMsg(
+    const std::vector<int> &params=std::vector<int>(), const Format dst_format = JPG) const;
```
christianrauch, line 137: Why do you need a second function of the same name with swapped parameter order?

cv_bridge/include/cv_bridge/cv_bridge.h
```diff
@@ -136,11 +150,25 @@ class CV_BRIDGE_EXPORT CvImage
    * can be: jpg, jp2, bmp, png, tif at the moment
    * support this format from opencv:
    * http://docs.opencv.org/modules/highgui/doc/reading_and_writing_images_and_video.html#Mat imread(const string& filename, int flags)
+   * params are the options for cv::imencode.
+   * Default value is empty list.
    */
   void toCompressedImageMsg(
     sensor_msgs::msg::CompressedImage & ros_image,
-    const Format dst_format = JPG) const;
+    const Format dst_format = JPG, const std::vector<int> &params=std::vector<int>()) const;
```
christianrauch, line 158: Use the curly braces initialiser (`{}`).

cv_bridge/src/cv_bridge.cpp
```diff
@@ -464,15 +464,21 @@ CvImagePtr cvtColor(
...
+  toCompressedImageMsg(*ptr, dst_format, params);
   return ptr;
 }
 
+sensor_msgs::msg::CompressedImage::SharedPtr CvImage::toCompressedImageMsg(const std::vector<int> &params, const Format dst_format)
+const
+{
+  return toCompressedImageMsg(dst_format, params); 
+}
+
```
christianrauch, line 481: What is this function actually doing? It just swaps the parameters.

cv_bridge/include/cv_bridge/cv_bridge.h
```diff
@@ -136,11 +150,25 @@ class CV_BRIDGE_EXPORT CvImage
...
+   * Default value is empty string that will convert to jpg format.
+   * can be: jpg, jp2, bmp, png, tif at the moment
+   * support this format from opencv:
+   * http://docs.opencv.org/modules/highgui/doc/reading_and_writing_images_and_video.html#Mat imread(const string& filename, int flags)
+   * params are the options for cv::imencode.
+   * Default value is empty list.
+   */
+  void toCompressedImageMsg(
+    sensor_msgs::msg::CompressedImage & ros_image,
+    const std::vector<int> &params=std::vector<int>(), const Format dst_format = JPG) const;
```
christianrauch, line 171: Here too. Why do we need the same function with swapped parameters?

christianrauch (changes requested):

cv_bridge/include/cv_bridge/cv_bridge.h
emrekuru97: I thought of this like, may be i want only provide parameters for the cv::imencode. Otherwise the dest_format need to be explicitly specified in each call we want provide params to cv::imencode.

cv_bridge/include/cv_bridge/cv_bridge.h
emrekuru97: Thanks for the detailed feedback. I'll let a function with the original signature. I thought since the parameter is default it will not alter the other packages i did not thought of compiling.

cv_bridge/src/cv_bridge.cpp
emrekuru97: That's only the implementation for providing parameters to cv::imencode without explicitly providing the format.

cv_bridge/include/cv_bridge/cv_bridge.h
emrekuru97: The same reason like the other function. If this is not neccessary i can remove it (swapping the parameters) and only provide cv::imencode params alongside with dst_format.

emrekuru97:
Hello again,
How should i proceed? With swapping(support for passing cv::imencode params without explicitly specifying dst_format) or without? 
Best regards

ijnek:
Hey @emrekuru97, thanks for opening this PR, and thanks for reviewing @christianrauch!

Despite the method "toCompressedImageMsg" existing in this package, moving forward, I don't think we should be extending cv_bridge's capability to compress images. cv_bridge's responsibility is to convert ROS Image msg <-> OpenCV image.

The thing that immediately came to mind when I read this was the compressed transport plugin, which is used to transport compressed image messages over a topic. In that plugin, the [conversion takes place, and then the compression is called afterwards](https://github.com/ros-perception/image_transport_plugins/blob/7404d772b66f586f5f56e568671ae20e8a64bc7e/compressed_image_transport/src/compressed_publisher.cpp#L222-L225). Would this work for your use case? (or could you just use the compressed publisher?

PR #534, opened 2024-06-12
add opencv_include_dirs in include

knorth55:
because we use `opencv` in include files, we need to set `OpenCV_INCLUDE_DIRS` in `INCLUDE_DIRS`.
https://github.com/ros-perception/vision_opencv/blob/a34a0c261984e4ab71f267d093f6a48820801d80/image_geometry/include/image_geometry/pinhole_camera_model.hpp#L7-L9


ijnek:
Thanks for the contribution!

PR #540, opened 2024-11-16
ensure same format encoding happens for all encoded types

### LINKED ISSUE #532: Magic number in `getConversionCodes()`
There is a magic 5 in [`getConversionCodes()`](https://github.com/ros-perception/vision_opencv/blob/a34a0c261984e4ab71f267d093f6a48820801d80/cv_bridge/src/cv_bridge.cpp#L154). Shouldn't it actually be 10, the last index of the `Encoding` enum?

Anyway, a comment should be added explaining the number there.

ijnek:
Resolves: #532


ijnek:
@Mergifyio backport humble

PR #549, opened 2025-05-30
Consider 8UC1 and 16UC1 as GRAY

galou:


ijnek:
We can't assume 8UC1 will be a grayscale image, anyone can store any type of data in ``8UC1``.

* If you fill this with intensity values from a camera: it's grayscale.
* If you fill it with binary mask values: it's a mask.
* If you fill it with per-pixel labels: it's a segmentation map.

galou:
Ok, thanks for the explanations.

PR #551, opened 2025-07-14
fix cmake deprecation

mosfet80:
cmake version < the 3.10 is deprecated
allign cmake version to 3.14. cv_bridge use -> cmake_minimum_required(VERSION 3.14)


mosfet80:
ping  @mjcarroll

clalancette (approved):

PR #544, opened 2025-02-10
Add support for NV12 encoding in cv_bridge

zycczy:
### Description

This PR adds support for the NV12 encoding in the `cv_bridge` package.  Change depends on common_interfaces PR: https://github.com/ros2/common_interfaces/pull/264

The following changes have been made:

- Added NV12 encoding to the list of supported encodings.
- Implemented conversion codes for NV12 to GRAY, RGB, BGR, RGBA, and BGRA.
- Updated `getCvType` and `getEncoding` functions to handle NV12 encoding.
- Modified `matFromImage` to correctly handle height scaling for NV12 encoding.

### Changes

1. **Encoding Support**:
    - Added `NV12` to the `Encoding` enum.
    - Updated `getCvType` and `getEncoding` functions to recognize `NV12`.

2. **Conversion Codes**:
    - Added conversion codes for `NV12` to `GRAY`, `RGB`, `BGR`, `RGBA`, and `BGRA` in `getConversionCodes`.

3. **Height Scaling**:
    - Modified `matFromImage` to account for height scaling when the encoding is `NV12`.

PR #552, opened 2025-08-04
Add support for NV12 encoding in cv_bridge

quic-zhanlin:
Base on #544 add support in utest2.cpp


quic-zhanlin:
@ijnek could you please have a review for this PR?

PR #555, opened 2025-10-21
prefix compressed 'format' with OpenCV type

### LINKED ISSUE #250: wrong "format" for compressed 16bit png images
Creating a `CompressedImage` in Python via:
```Python
img_msg = self.cvbridge.cv2_to_compressed_imgmsg(img, dst_format="png")
```
where `img` is is `np.uint16`, creates a compressed image message with `img_msg.format = "png"`, i.e. it sets `format` directly to `dst_format`:
https://github.com/ros-perception/vision_opencv/blob/a3736d24554166142981fa2bcf3d7d3c6e3ce366/cv_bridge/python/cv_bridge/core.py#L216
However, `rqt_image_view` only displays the compressed 16bit png image properly when the format is set to:
```Python
img_msg.format = "16UC1; png compressed"
```

christian-rauch:
Fixes https://github.com/ros-perception/vision_opencv/issues/250.

### DIFF
--- a/cv_bridge/python/cv_bridge/core.py
+++ b/cv_bridge/python/cv_bridge/core.py
@@ -227,7 +227,11 @@ class CvBridge(object):
         if not isinstance(cvim, (np.ndarray, np.generic)):
             raise TypeError('Your input type is not a numpy array')
         cmprs_img_msg = sensor_msgs.msg.CompressedImage()
-        cmprs_img_msg.format = dst_format
+        if len(cvim.shape) < 3:
+            cv_type = self.dtype_with_channels_to_cvtype2(cvim.dtype, 1)
+        else:
+            cv_type = self.dtype_with_channels_to_cvtype2(cvim.dtype, cvim.shape[2])
+        cmprs_img_msg.format = f"{cv_type}; {dst_format} compressed"
         ext_format = '.' + dst_format
         try:
             cmprs_img_msg.data.frombytes(np.array(cv2.imencode(ext_format, cvim)[1]).tobytes())


christian-rauch:
@clalancette @mjcarroll I am pinging you, as this repo seems unmaintained, and you are the last persons approving or merging PRs.

Can you have a look at my PR?

There are a couple of approve but unmerged PRs. If you need help with this, and I have contributed a couple of PRs before, can you make me a maintainer of this and the other related vision repos?

christian-rauch:
@clalancette @mjcarroll Can you please have a look at this, tell me who is maintaining this, or make me a maintainer, so these issues can be fixed?

fallenmi (changes requested):
Disclosure: I used Codex to assist with this review and verified the finding against exact commit `0389e888100f92e1c4590314b4ea45f05a8ca321`.

**[P1] Do not advertise a source depth that `imencode()` did not preserve**

This builds the `CompressedImage.format` prefix from the input array before calling `cv2.imencode()`. PNG preserves the intended `uint8` and `uint16` cases, but OpenCV falls back to `CV_8U` for the other NumPy depths accepted by this public method. The resulting metadata then disagrees with the encoded payload.

For example, exact head accepts a four-column `float32` mono image, emits `format='32FC1; png compressed'`, and writes a PNG that decodes as `8UC1`. The current `compressed_image_transport` subscriber copies the prefix into `Image.encoding`, but builds `step` from the decoded matrix, so it publishes `encoding=32FC1, step=4`; `cv_bridge::matFromImage()` requires at least 16 bytes for that width and rejects the message as malformed. A three-channel `int16` input similarly becomes `encoding=16SC3, step=12` where 24 bytes are required.

I exercised PNG encoding and the exact subscriber format-selection contract across seven NumPy depths and one/three channels on the PR base, exact head, and current GitHub merge. The base exposes the existing two `uint16` metadata failures that this PR is intended to fix. The head/merge fix both of those, but introduce ten invalid signed/float combinations (`int8`, `int16`, `int32`, `float32`, and `float64`, each at one and three channels); all three matrices reproduced 10/10. The `uint8` and `uint16` controls remain correct on head.

Please either reject/convert unsupported input depths before constructing the metadata or derive the advertised type from what was actually encoded, and add signed/float controls alongside the `uint16` regression. `git diff --check` and Python compilation pass; this PR has no attached Actions run, only the two completed Mergify statuses.

PR #556, opened 2025-11-17
Deprecate boost and use nanobind for the python binding

Benjamin-Tan:
This PR uses the modern and simplified Python binding (nanobind) over boost, which allows the Python bindings to work seamlessly with both NumPy 1.x and NumPy 2.x. #535


Benjamin-Tan:
@ijnek do you have some time to have a look at this? Or know of anyone who is suitable to review this PR.

cv_bridge/python/cv_bridge/core.py
```diff
@@ -188,7 +188,7 @@ def imgmsg_to_cv2(self, img_msg, desired_encoding='passthrough'):
         if desired_encoding == 'passthrough':
```
ziadamr22, line 188: For the cv_bridge to support numpy >=2.0, line 186 need to be edited to 
`im = im.byteswap().view(im.dtype.newbyteorder())`
as .newbyteorder() is not supported in numpy>=2.0, which caused errors when using imgmsg_to_cv2

ziadamr22 (changes requested):

ignacio-pm:
Will this PR be merged also to humble?

dvdmc:
Any update on this? Will this be backported to Jazzy / Humble?

PR #275, opened 2019-05-10
Fix final confusion between format and encoding.

talregev:
Fix final confusion between format and encoding.
Format is file format like jpg, bmp and more,
encoding is color encoding like bgr8, rgb8 and more.

PR #562, opened 2025-12-18
Fix final confusion between format and encoding.

talregev:
Format is file format like jpg, bmp and more,
encoding is color encoding like bgr8, rgb8 and more.

continue of #275

### DIFF
--- a/cv_bridge/src/cv_bridge.cpp
+++ b/cv_bridge/src/cv_bridge.cpp
@@ -142,17 +142,17 @@ Encoding getEncoding(const std::string & encoding)
   return INVALID;
 }
 
-static const int SAME_FORMAT = -1;
+static const int SAME_ENCODING = -1;
 
-/** Return a lit of OpenCV conversion codes to get from one Format to the other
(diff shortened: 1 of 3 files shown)


talregev:
@clalancette FYI

cv_bridge/src/cv_bridge.cpp
```diff
@@ -142,17 +142,17 @@ Encoding getEncoding(const std::string & encoding)
   return INVALID;
 }
 
-static const int SAME_FORMAT = -1;
+static const int SAME_ENCODING = -1;
 
-/** Return a lit of OpenCV conversion codes to get from one Format to the other
- * The key is a pair: <FromFormat, ToFormat> and the value a succession of OpenCV code conversion
+/** Return a lit of OpenCV conversion codes to get from one encoding to the other
```
clalancette: While we are here:

```suggestion
/** Return a list of OpenCV conversion codes to get from one encoding to the other
```

clalancette (changes requested):

clalancette (approved):

PR #564, opened 2026-04-10
cuda demosaicing if available

simonblaue:
I added the cuda demosaicing from opencv, if cuda is found by cmake. For this to work, you need to have a custom compiled opencv with cuda for cmake to find.

### DIFF
--- a/cv_bridge/CMakeLists.txt
+++ b/cv_bridge/CMakeLists.txt
@@ -48,8 +48,23 @@ find_package(OpenCV 4 QUIET
     opencv_core
     opencv_imgproc
     opencv_imgcodecs
+    opencv_cudaimgproc
   CONFIG
 )
+
+if(OpenCV_FOUND)
+add_compile_definitions("HAVE_OPENCV_CUDAIMGPROC")
+set(WITH_OPENCV_CUDAIMGPROC TRUE)
+else()
+  find_package(OpenCV 4 QUIET
+    COMPONENTS
+      opencv_core
+      opencv_imgproc
+      opencv_imgcodecs
+    CONFIG
+  )
+endif()
+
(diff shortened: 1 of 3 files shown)


fallenmi (changes requested):
Requesting changes for the no-device fallback path.

When `opencv_cudaimgproc` is available at build time, `getConversionCodes()` replaces every Bayer conversion code with the CUDA MHT values (`256`-`263`). If the same binary runs without a visible CUDA device, however, `toCvCopyImpl()` falls back to CPU `cv::cvtColor()` while keeping those CUDA-only codes. The CPU color-conversion enum ends at `155`, so Bayer conversion throws instead of falling back.

I reproduced this against exact head `f50e677bc24d8e634269d8ff2328289708af8575` with an actual OpenCV 4.15 core/imgproc runtime and the actual cudaimgproc enum definitions. With `getCudaEnabledDeviceCount() == 0`, all 12 base CPU Bayer-to-gray/RGB/BGR codes succeed, while all 12 codes selected by this head throw OpenCV error `-206`; the asserted output was byte-identical across ten runs. The current GitHub merge has the same tree as the head.

Please keep a CPU conversion code for the zero-device branch (or select the MHT code only inside the CUDA-device branch), and add a regression for a build that has `opencv_cudaimgproc` but runs with no visible GPU.

The three retained ROS build statuses are green. I could not run the full repository suite locally because this host does not have the ROS/colcon toolchain; the focused oracle compiles warning-clean against the exact OpenCV APIs used by the changed paths.

Disclosure: I used Codex to inspect the exact diff and build/runtime branches, construct and run the OpenCV fallback oracle, check the live refs and public policy, and draft this review. I verified the source lines, enum values, outputs, commit IDs, and conclusion.

PR #565, opened 2026-04-19
Replace rcpputils::endian with std::endian

mini-1235:
Related to https://github.com/ros2/rcpputils/pull/227


mini-1235:
this PR also bumps all packages to c++20

